PR #2, opened 2015-02-26
DSpace updates and re-run updated transfers (deprecated)

Hwesta:
- Add DSpace specific pre-transfer metadata processing
- Re-run transfers in transfer source that have been updated.  This does not currently handle timezones.

Supporting, now in #3 and #4:
- Switch to using argparse instead of docopt because argparse does input validation
- Support running with Python2 or Python3
- Add and update tests
- Add support for different transfer types

### DIFF
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -342,7 +342,7 @@ def start_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, ts_path, depth
     """
     # Start new transfer
     completed = {x[0] for x in session.query(models.Unit.path).all()}
-    transfer_start_times = session.query(Unit.path, Unit.started_timestamp)
+    transfer_start_times = session.query(models.Unit.path, models.Unit.started_timestamp)
     target = get_next_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, ts_path, depth, completed, see_files, transfer_start_times)
     if not target:
         LOGGER.warning("All potential transfers in %s have been created. Exiting", ts_path)
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -10,6 +10,7 @@ import argparse
 import ast
 import base64
 import datetime
+from dateutil.parser import parse
 import logging
 import logging.config  # Has to be imported separately
 import os
@@ -226,7 +227,7 @@ def run_scripts(directory, *args):
             LOGGER.warning('stderr: %s', stderr)
 
 
-def get_next_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, path_prefix, depth, completed, see_files):
+def get_next_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, path_prefix, depth, completed, see_files, transfer_start_times=None):
     """
     Helper to find the first directory that doesn't have an associated transfer.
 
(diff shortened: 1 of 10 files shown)


transfers/get-accession-number
```diff
@@ -7,9 +7,9 @@ def main(dirname):
     # Expecting a directory name like sometext---dc.identifier---accession
     parts = dirname.rsplit('---', 1)
     try:
-        print('"' + parts[1] + '"')  # Accession ID must be quoted
+        print('"' + parts[1] + '"', end='')  # Accession ID must be quoted
```
mistydemeo: What's the reason for this change?

transfers/pre-transfer/defaultProcessingMCP.xml
```diff
@@ -68,5 +68,9 @@
       <appliesTo>b320ce81-9982-408a-9502-097d0daa48fa</appliesTo>
       <goToChain>/api/v2/location/edc821af-b9c0-480f-9465-e6b38b23dc99/</goToChain>
     </preconfiguredChoice>
+    <preconfiguredChoice>
+      <appliesTo>05f99ffd-abf2-4f5a-9ec8-f80a59967b89</appliesTo>
```
mistydemeo: Add a comment about which one this is?

transfers/transfer.py
```diff
@@ -189,8 +200,11 @@ def get_next_transfer(ss_url, ts_location_uuid, path_prefix, depth, completed):
     browse_info = _call_url_json(url, params)
     if browse_info is None:
         return None
-    dirs = browse_info['directories']
-    dirs = map(base64.b64decode, dirs)
+    if see_files:
+        dirs = browse_info['entries']
```
mistydemeo: Extreme nitpick, but since this is no longer always directories, maybe worth renaming the variable?

mistydemeo:
Looks good, aside from comments!

transfers/get-accession-number
Hwesta: Thought it looked nicer?  I think I was debugging ast.literal_eval stuff, but this doesn't seem to affect it.

transfers/transfer.py
Hwesta: Haha, I was thinking of doing that but didn't to keep the diff small. Now I have an excuse.

transfers/get-accession-number
mistydemeo: Since this is being printed, my preference'd be to print with a newline; it's standard for the output of CLI tools to end with a newline. (POSIX actually [defines](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) a "line" of text as ending with a newline; some commandline tools in pipelines may act strangely if data doesn't end with a newline.

Hwesta:
Broke this PR into two others: #3 for the general python & argparse updates and #4 for the transfer creation updates.

hakamine:
As mentioned in the first message in this conversation, this PR includes the following features:

1) Add a pre-transfer script for DSpace specific metadata processing

2) Re-run transfers in transfer source that have been updated (this does not currently handle timezones)

Feature 2) does not seem to work well with the DSpace packager tool used to create exports (the exporter is documented in https://wiki.duraspace.org/display/DSDOC6x/AIP+Backup+and+Restore#AIPBackupandRestore-ExportingAIPs and https://wiki.duraspace.org/display/DSDOC6x/Importing+and+Exporting+Content+via+Packages). Each time the DSpace exporter runs, it exports everything, even if the source content is unchanged, meaning that the date stamps on the created ZIP files also change every time it's run. If the  commits for 2) are merged, it will potentially cause the automation tools to ingest all the already completed transfer sources again every time the DSpace exporter runs, which is not desired. 

I created branch dev/dspace-merge (PR #49) that includes only the commit related to 1) which should be ready to merge. I will also rename this PR as "DSpace updates and re-run updated transfers (deprecated)" in order to avoid the risk of unintentionally merging it in its current state.

hakamine:
Closing this without merging (see previous comment)

PR #3, opened 2015-03-30
Support Python3, use argparse

Hwesta:
* Switch to using argparse instead of docopt because argparse does input validation
* Support running with Python2 or Python3
* Add and update tests

### DIFF
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -1,4 +1,4 @@
 -r base.txt
-ipython
+ipython==1.0.0
 pytest
 vcrpy>=1.0.0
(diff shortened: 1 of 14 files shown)


requirements/local.txt
```diff
@@ -1,4 +1,4 @@
 -r base.txt
-ipython
+ipython==1.0.0
```
mistydemeo: Why fix this to 1.0?

Hwesta: To work with Python 3.2, which is what Ubuntu 12.04 ships for python 3.  "IPython supports Python 2.7 and 3.3 or newer. Our older 1.x series supports Python 2.6 and 3.2." :(

mistydemeo:
:+1:

If this is being ported to Python 3, maybe use subprocess32 in place of subprocess on Python 2.x?

Hwesta:
I'd like 3 to be the default, and 2 to be backwards compatibility. :) We don't use subprocess for much, so I think I'll leave it as is.  Thanks for letting me know about that library.

PR #4, opened 2015-03-30
Support different transfer types, transfers from files

Hwesta:
* Specify transfer type as a parameter
* Allow creating transfers from files
* Add script to move files to a directory of the same name, to support other pre-transfer microservices (eg adding processingMCP)


mistydemeo:
:+1:

PR #7, opened 2015-07-10
Pre-transfer hook for ArchivesSpace IDs

Hwesta:
Pre-transfer hook to generate ArchivesSpace IDs from the filename and writes to an archivesspaceids.csv

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ There are some sample scripts in the pre-transfers directory that may be useful,
 
 * `00_file_to_folder.py`: If the transfer is a single file (eg a zipped bag or DSpace transfer), it moves it into an identically named folder. This is not required for processing, but allows other pre-transfer scripts to run.
 * `add_metadata.py`: Creates a metadata.json file, by parsing data out of the transfer folder name.  This ends up as Dublin Dore in a dmdSec of the final METS file.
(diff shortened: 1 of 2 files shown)


transfers/pre-transfer/archivesspace_ids.py
```diff
@@ -0,0 +1,40 @@
...
+    try:
+        os.mkdir(os.path.join(transfer_path, 'metadata'))
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            pass  # Already exists
+        else:
+            raise
+    with open(os.path.join(transfer_path, 'metadata', 'archivesspaceids.csv'), 'w') as f:
+        writer = csv.writer(f)
+        for row in as_ids:
```
mistydemeo: How about:

```python
writer.writerows(as_ids)
```

mistydemeo:
:+1: 

Is the lack of a header row consistent with the other tools?

transfers/pre-transfer/archivesspace_ids.py
Hwesta: Good catch. Fixed.

hakamine:
Would it be ok to merge to master once #20 is merged to dev/archivesspace ?

Hwesta:
That seems reasonable.

PR #8, opened 2015-10-15
Adapt transfers with checksums per file

sevein:
See the commit message for more details.

### DIFF
--- a/transfers/pre-transfer/fix_unlv_checksums.py
+++ b/transfers/pre-transfer/fix_unlv_checksums.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+
+import sys
+import os
+
+
+HASHFUNC_EXT = 'md5'
+
+
+def extract_checksum(filename, metadata_path):
+    filename = '{}.{}'.format(os.path.join(metadata_path, filename),
+                              HASHFUNC_EXT)
+    try:
+        with open(filename, 'r') as f:
+            return f.readline().strip()
+    except IOError:
+        return None
+
+
+def main(transfer_path):
+    """
+    Generate the standard checksum file in a transfer where every object has
+    its checksum stored in a separate text file under the metadata directory,
+    e.g. transfer-A is the original transfer and transfer-B is the resulting
+    transfer after running this script.
+
+       .
+       ├── transfer-A
+       |   ├── foobar.mp3
+       │   └── metadata
+       │       └── foobar.mp3.md5
+       └── transfer-B
+        ├── foobar.mp3
+        └── metadata
+            ├── checksum.md5
+            └── foobar.mp3.md5
+
+    """
+    transfer_path = os.path.abspath(transfer_path)
+    metadata_path = os.path.join(transfer_path, 'metadata')
+
+    if not os.path.exists(metadata_path):
+        print('metadata directory cannot be found', metadata_path,
+              file=sys.stderr)
+        return 1
+
+    checksum_file = os.path.join(metadata_path, 'checksum.{}'.format(
+                                 HASHFUNC_EXT))
+    with open(checksum_file, 'w') as f:
+        for root, dirs, files in os.walk(transfer_path):
+            if root == transfer_path:
+                dirs.remove('metadata')
+            for item in files:
+                file_path = os.path.join(root, item)
+                hashsum = extract_checksum(item, metadata_path)
+                if hashsum is None:
+                    print('Cannot find checksum of', file_path, file=sys.stderr)
+                    continue
+                hashfile = './{}'.format(os.path.relpath(file_path, transfer_path))
+                checksum_entry = '{}  {}\n'.format(hashsum, hashfile)
+                f.write(checksum_entry)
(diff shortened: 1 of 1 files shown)


transfers/pre-transfer/fix_unlv_checksums.py
```diff
@@ -0,0 +1,83 @@
...
+            basename = os.path.basename(item)
+            hashsum = extract_checksum(basename, metadata_path)
+            if hashsum is None:
+                print('Cannot find checksum of {}'.format(basename),
+                      file=sys.stderr)
+            hashfile = './{}'.format(os.path.relpath(item, transfer_path))
+            checksum_entry = '{}  {}\n'.format(hashsum, hashfile)
+            f.write(checksum_entry)
+
+    print('Checksums file generated: {}'.format(checksum_file))
```
Hwesta: Can be just `print('Checksums file generated:', checksum_file)`

transfers/pre-transfer/fix_unlv_checksums.py
```diff
@@ -0,0 +1,83 @@
...
+
+    checksum_file = os.path.join(metadata_path, 'checksum.{}'.format(
+                                 HASHFUNC_EXT))
+    with open(checksum_file, 'w') as f:
+        for item in find_objects(transfer_path, ignore_dirs=metadata_path):
+            basename = os.path.basename(item)
+            hashsum = extract_checksum(basename, metadata_path)
+            if hashsum is None:
+                print('Cannot find checksum of {}'.format(basename),
+                      file=sys.stderr)
```
Hwesta: `continue` after this?

Print knows how to convert things to strings, so calling .format on them isn't needed.  This could be `print('Cannot find checksum of', basename, file=sys.stderr)`

transfers/pre-transfer/fix_unlv_checksums.py
```diff
@@ -0,0 +1,83 @@
...
+from __future__ import print_function
+
+import sys
+import os
+
+
+HASHFUNC_EXT = 'md5'
+
+
+def find_objects(path, ignore_dirs=None):
```
Hwesta: Are you familiar with [`os.walk`](https://docs.python.org/2/library/os.html#os.walk)?  This looks like a reimplementation of it.

transfers/pre-transfer/fix_unlv_checksums.py
```diff
@@ -0,0 +1,83 @@
...
+        elif os.path.isfile(full_path):
+                yield full_path
+
+
+def extract_checksum(filename, metadata_path):
+    filename = '{}.{}'.format(os.path.join(metadata_path, filename),
+                              HASHFUNC_EXT)
+    checksum = None
+    try:
+        with open(os.path.join(metadata_path, filename), 'r') as f:
```
Hwesta: Looks like `metadata_path` is being added twice - here and 4 lines above.

transfers/pre-transfer/fix_unlv_checksums.py
```diff
@@ -0,0 +1,83 @@
...
+            checksum = f.readline().strip()
+    except:
+        pass
```
Hwesta: `except: pass` is generally considered a bad idea because it silences *everything* including normal-workings exceptions.  `except Exception: pass` is better, listing the exceptions you expect is better.

transfers/pre-transfer/fix_unlv_checksums.py
sevein: Yes I am! I thought this was nicer because I'm decoupling the process of listing the files in the way I want from its consumer. Does that make sense? I'm happy to use `os.walk` if that's the practice.

transfers/pre-transfer/fix_unlv_checksums.py
sevein: True!

transfers/pre-transfer/fix_unlv_checksums.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if not os.path.exists(metadata_path):
+        print('metadata directory cannot be found: {}'.format(metadata_path),
+              file=sys.stderr)
+        return 1
+
+    checksum_file = os.path.join(metadata_path, 'checksum.{}'.format(
+                                 HASHFUNC_EXT))
+    with open(checksum_file, 'w') as f:
+        for item in find_objects(transfer_path, ignore_dirs=metadata_path):
```
Hwesta: An incoming transfer can have several 'magic' directories: metadata, logs, manualNormalization, access, service.  You could try to ignore all the magic directories, or just assume files in them don't have checksum so they'll be a noop later?

transfers/pre-transfer/fix_unlv_checksums.py
Hwesta: Why not use os.walk and check if it matches/startswith the ignore_dirs (ignore_paths?) before yielding the value?

sevein: :+1:

Hwesta:
:+1:

Hwesta:
This needs to be updated for #12

PR #9, opened 2016-01-20
Automate Transfers: Fix 'user' param to 'username'

Hwesta:
API auth actually checks the 'username' parameter, not 'user'.


mistydemeo:
:+1:

PR #10, opened 2016-01-21
Automate Transfers: Hide units when complete

Hwesta:
Add a `--hide` switch that will hide the Transfers/SIPs in Archivematica when they complete.  It will not hide units that are rejected or failed, but it will hide units that ran to completion with errors.


finoradin:
Nice!!

Hwesta:
Note: This relies on artefactual/archivematica#386

mistydemeo:
Maybe make `hide` a keyword argument? I find boolean positional arguments to be unclear.

Hwesta:
Moved 'hide' to the end with a default of False.

mistydemeo:
:+1:

PR #1, opened 2015-01-16
Add script to create JSON file with metadata for NetX upload refs #7772

hakamine:
Add a script that creates a metadata file called netx.json,
that will contain the ComponentNumber information. This file
is used by the CopyToNetX.py archivematica client script.

PR #12, opened 2016-02-18
Automate Transfers: Move sample hook scripts to examples directory

Hwesta:
Move example scripts to an examples directory, update docs to explain change.


Hwesta:
PRs #1 #2 #7 and #8 should be updated to put the client script in the examples directory.

jhsimpson:
:+1: 

Does this PR invalidate or change any of the existing open PR's?  Should they be rebased on top of this?

hakamine:
:+1:   ( I think it would also be great to move some of the user configuration to /etc/archivematica/automation-tools/ for example, and make the scripts fetch from there, maybe in a future PR? )

PR #15, opened 2016-03-21
Pre-transfer hooks for Dataverse

Hwesta:
Parse dataset.json and produce a METS.xml with the DDI info and a structMap with the expected transfer structure inside the research data package.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,2 +1,4 @@
 requests<3.0
 sqlalchemy
+lxml
+-e git://github.com/artefactual-labs/mets-reader-writer.git@dev/dataverse#egg=metsrw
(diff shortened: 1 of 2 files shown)


ross-spencer:
@Hwesta I don't know if you have been following the progress on this, but we merged your excellent work here into Archivematica mainline and it will be available with the `1.8` release: https://github.com/artefactual/archivematica/commit/06c0b31f7917918cd878d5cdf4c1fb10aea466ee 

I had a lot of learning to do on other parts of the system. I think aside from a few CR-like changes and a little re-factoring, much of your code remains the same as you created it, so thank you! 

I think we have a few folks running Dataverse aside from OCUL so I am excited to see where this integration goes.

PR #16, opened 2016-04-14
Automate transfers: handle no accession id script

Hwesta:
Since there's no default get-accession-id script anymore, this shouldn't error if one doesn't exist.

### DIFF
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -156,7 +156,11 @@ def get_accession_id(dirname):
     :returns: accession number or None.
     """
     script_path = os.path.join(THIS_DIR, 'get-accession-number')
-    p = subprocess.Popen([script_path, dirname], stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    try:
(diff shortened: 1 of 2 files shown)


transfers/transfer.py
```diff
@@ -296,7 +300,7 @@ def start_transfer(ss_url, ts_location_uuid, ts_path, depth, am_url, user_name,
         return None
     if not response.ok or resp_json.get('error'):
         LOGGER.error('Unable to start transfer.')
-        LOGGER.debug('Response: %s', resp_json)
+        LOGGER.error('Response: %s', resp_json)
```
sevein: Should the level of this message be `DEBUG`?

sevein:
👍

transfers/transfer.py
Hwesta: It was, but any time you see that message you have an error, and it's always useful to be able to see the erroneous response (and not have to look for DEBUG logging)

sevein: We should probably have just one single log debug stream in development environments, but we can look at that later! Thanks!

PR #18, opened 2016-04-22
transfers: add config file for log/db/pid file paths

hakamine:
Avoid creating log/db/pid files in the code path:
- Make models.py and transfer.py find config information in file
  /etc/archivematica/automation-tools/automation-tools.conf
- Provide example config files in etc/ directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,37 +3,54 @@ Automation Tools
 
 The Automation Tools project is a set of python scripts, that are designed to automate the processing of transfers in an Archivematica pipeline.
 
-Currently, the only automation tool is automate transfers.  It is used to prepare transfers, move them into the pipelines processing location, and take actions when user input is required.  Only one transfer is sent to the pipeline at a time, the scripts wait until the current transfer is resolved (failed, rejected or stored as an AIP) before automatically starting the next available transfer. 
+Installation
+------------
 
-The code is available on [Github](http://github.com/artefactual/automation-tools).
+* Checkout or link the code in this repo to `/usr/lib/archivematica/automation-tools`
+* Create virtualenv `/usr/share/python/automation-tools` and pip install requirements there
+* Create directories `/var/log/archivematica/automation-tools` and `/var/archivematica/automation-tools` owned by user `archivematica`, for log/database/PID files.
+* Create directory `/etc/archivematica/automation-tools` and add configuration files there. Files in the `etc/` directory of this repository can be used as an example (also see below for more about configuration)
 
-The code is deployed to `/usr/lib/archivematica/automation-tools`.
+Automated transfers
+-------------------
 
-Deployment
-----------
+`transfers/transfer.py` is used to prepare transfers, move them into the pipelines processing location, and take actions when user input is required.
+Only one transfer is sent to the pipeline at a time, the scripts wait until the current transfer is resolved (failed, rejected or stored as an AIP) before automatically starting the next available transfer.
 
-Suggested deployment is to use cron to run a shell script that runs the automate transfer tool. Example shell script:
+### Configuration - Overview
 
-    #!/bin/bash
-    cd /usr/lib/archivematica/automation-tools/transfers/
-    /usr/share/python/automation-tools/bin/python transfer.py --user <add user>  --api-key <add api key> --transfer-source <add transfer source location uuid> --depth 2
+Suggested deployment is to use cron to run a shell script that runs the automate transfer tool. Example shell script (for example in `/etc/archivematica/automation-tools/transfer-script.sh`):
 
-This script is run through a crontab entry. Example:
+```
+#!/bin/bash
+cd /usr/lib/archivematica/automation-tools/
+/usr/share/python/automation-tools/bin/python -m transfers.transfer --user <user>  --api-key <apikey> --transfer-source <transfer_source_uuid> --config-file <config_file>
+```
 
-    */5 * * * * /etc/archivematica/automation-tools/transfer-script.sh
+(Note that the script calls the transfers script as a module using python's `-m` flag, this is required due to the use of relative imports in the code)
 
-The cron entry executes the `transfer-script.sh` script. This should be run as the same user as Archivematica is run as (likely the `archivematica` user.)
+The script can be run from a shell window like:
 
-When running, automate transfers stores its working state in transfers.db, a sqlite database.  It contains a record of all the transfers that have been processed.  In a testing environment, deleting this file will cause the tools to re-process any and all folders found in the Transfer Source Location. 
+```
+user@host:/etc/archivematica/automation-tools$ sudo -u archivematica ./transfer-script.sh
+```
 
-Configuration
--------------
+It is suggested to run the script through a crontab entry for user archivematica (to avoid the need to repeatedly invoke it manually):
 
-This script can be modified to adjust how automated transfers work.  The full set of parameters that can be changed are:
+```
+*/5 * * * * /etc/archivematica/automation-tools/transfer-script.sh
+```
+
+When running, automated transfers stores its working state in a sqlite database.  It contains a record of all the transfers that have been processed.  In a testing environment, deleting this file will cause the tools to re-process any and all folders found in the Transfer Source Location.
+
+### Configuration - Parameters
+
+The `transfers.py` script can be modified to adjust how automated transfers work.  The full set of parameters that can be changed are:
 
 * `-u USERNAME, --user USERNAME` [REQUIRED]: Username of the dashboard user to authenticate as.
 * `-k KEY, --api-key KEY` [REQUIRED]: API key of the dashboard user.
 * `-t UUID, --transfer-source UUID`: [REQUIRED] Transfer Source Location UUID to fetch transfers from. Check the next section for more details on this field.
+* `-c FILE, --config-file FILE`: config file containing file paths for log/database/PID files. Default: log/database/PID files stored in the same directory as the script (not recommended for production)
 * `--transfer-path PATH`: Relative path within the Transfer Source. Default: ""
 * `--depth DEPTH, -d DEPTH`: Depth to create the transfers from relative to the transfer source location and path. Default of 1 creates transfers from the children of transfer-path.
 * `--am-url URL, -a URL`:Archivematica URL. Default: http://127.0.0.1
@@ -48,16 +65,13 @@ The easiest way to configure the tasks that automation-tools will run is by usin
 
 1. Go to Administration|Processing Configuration and choose the options you wish to use.
 
-2. Save the configuation on the form.
+2. Save the configuration on the form.
 
 3. Copy the processing configuration file from '/var/archivematica/sharedDirectory/sharedMicroServiceTasksConfigs/processingMCPConfigs/defaultProcessingMCP.xml' on the Archivematica host machine to the transfers/ directory of your automation-tools installation location.
 
-
 The automation-tools command-line also relies on installation-specific UUIDs. To obtain the transfer source UUID for script invocation, visit the 'Transfer Source' tab in the Archivematica Storage Space web dashboard. If a row is marked as a transfer souce its UUID value will be valid as a transfer source argument.
 
-
-Hooks
------
+### Hooks
 
 During processing, automate transfers will run scripts from several places to customize behaviour. These scripts can be in any language. If they are written in Python, we recommend making them source compatible with python 2 or 3.
 
@@ -73,7 +87,7 @@ There are also several scripts provided for common use cases and examples of pro
 These are found in the `examples` directory sorted by their usecase and can be copied or symlinked to the appropriate directory for automation-tools to run them.
 If you write a script that might be useful for others, please make a pull request!
 
-### get-accession-id
+#### get-accession-id
 
 * _Name:_ `get-accession-id`
 * _Location:_ Same directory as transfers.py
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -3,11 +3,12 @@ Automation Tools
...
+Currently, the only automation tool is automate transfers.  It is used to prepare transfers, move them into the pipelines processing location, and take actions when user input is required.  Only one transfer is sent to the pipeline at a time, the scripts wait until the current transfer is resolved (failed, rejected or stored as an AIP) before automatically starting the next available transfer.
 
 The code is available on [Github](http://github.com/artefactual/automation-tools).
 
 The code is deployed to `/usr/lib/archivematica/automation-tools`.
+Configuration files should go into `/etc/archivematica/automation-tools/`, the automation tools require a configuration file named `automation-tools.conf` in this directory to run (a sample file is provided in the `etc/` directory of this repository).
```
Hwesta: Nitpick: I think this makes more sense split up more.  eg "Configuration files should go into `/etc/archivematica/automation-tools/`.  A configuration file named `automation-tools.conf` is required, and a sample is provided..."

transfers/transfer.py
```diff
@@ -16,6 +16,16 @@
...
 import time
+import ConfigParser
+
+# Read config file
+automationToolsConfigFilePath = '/etc/archivematica/automation-tools/automation-tools.conf'
+config = ConfigParser.SafeConfigParser()
+config.read(automationToolsConfigFilePath)
+logDirectory = config.get('automation-tools', 'logDirectory')
+databaseDirectory = config.get('automation-tools', 'databaseDirectory')
+pidFileDirectory = config.get('automation-tools', 'pidFileDirectory')
```
Hwesta: Do we want to specify a default value fo these if it's not found in the config file?

Hwesta:
A bit of minor feedback, otherwise :+1:

hakamine:
I think it's ready now. I had to move things around a little (such as putting the code of models.py inside a function, and taking ArgumentParser related code out of main in transfers.py) in order to allow the scripts to get config parameters from a config file specified in the command line. I did a few tests, seems to be working fine so far.
The README has also been updated, added new information on installation, and how to set up multiple instances.

README.md
```diff
@@ -3,37 +3,52 @@ Automation Tools
...
+When running, automated transfers stores its working state in a sqlite database.  It contains a record of all the transfers that have been processed.  In a testing environment, deleting this file will cause the tools to re-process any and all folders found in the Transfer Source Location.
+
+### Configuration - Parameters
+
+The `transfers.py` script can be modified to adjust how automated transfers work.  The full set of parameters that can be changed are:
 
 * `-u USERNAME, --user USERNAME` [REQUIRED]: Username of the dashboard user to authenticate as.
 * `-k KEY, --api-key KEY` [REQUIRED]: API key of the dashboard user.
 * `-t UUID, --transfer-source UUID`: [REQUIRED] Transfer Source Location UUID to fetch transfers from. Check the next section for more details on this field.
+* `-c FILE, --config-file FILE`: [OPTIONAL] config file containing locations and file names for log/database/PID files. If not specified default values are used.
```
Hwesta: I had only added [REQUIRED] to required options - I didn't mark optional things. Remove [OPTIONAL] or add it to the rest of the options. Also, what's the default behaviour?

README.md
```diff
@@ -48,16 +63,13 @@ The easiest way to configure the tasks that automation-tools will run is by usin
 
 1. Go to Administration|Processing Configuration and choose the options you wish to use.
 
-2. Save the configuation on the form.
+2. Save the configuration on the form.
```
Hwesta: Good catch.

README.md
```diff
@@ -132,6 +143,22 @@ Logs are written to the same directory as the `transfers.py` script. The logging
...
+### Multiple automated transfer instances
+
+You may need to set up multiple automated transfer instances, for example if required to ingest both standard transfers and bags. In cases where hooks are the same for both instances, it could be achieved by setting up different scripts, each one invoking the transfers.py script with the required parameters. Example:
+
+```
+# first script invokes like this (standard transfer):
+usr/share/python/automation-tools/bin/python transfer.py --user <user>  --api-key <apikey> --transfer-source <transfer_source_uuid_for_std_xfers> --config-file <config_file>
+
+# second script invokes like this (unzipped bags):
+usr/share/python/automation-tools/bin/python transfer.py --user <user>  --api-key <apikey> --transfer-source <transfer_source_2_uuid_for_bags> --config-file <config_file_2> --transfer-type 'unzipped bag'
```
Hwesta: Should this have a leading `/`?

transfers/transfer.py
```diff
@@ -16,6 +16,10 @@
 import subprocess
 import sys
 import time
+import ConfigParser
```
Hwesta: Python3 renamed this to [configparser](https://docs.python.org/3/library/configparser.html) Maybe we should add six as a dependency at this point? Or
```
try:
    import ConfigParser as configparser
except ImportError:
    import configparser
```

transfers/models.py
```diff
@@ -5,24 +5,32 @@
 from sqlalchemy import Column, Binary, Boolean, Integer, String
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
+import ConfigParser
```
Hwesta: Unused

transfers/transfer.py
```diff
@@ -419,13 +458,13 @@ def main(user, api_key, ts_uuid, ts_path, depth, am_url, ss_url, transfer_type,
...
-            status_info['name'],  # SIP/Transfer name
-            status_info['type'],  # SIP or transfer
-        )
+                    microservice,  # Current microservice name
+                    str(microservice != current_unit.microservice),  # String True or False if this is the first time at this wait point
+                    status_info['path'],  # Absolute path
+                    status_info['uuid'],  # SIP/Transfer UUID
+                    status_info['name'],  # SIP/Transfer name
+                    status_info['type'],  # SIP or transfer
+                    )
```
Hwesta: More of a style/preference thing, but I think moving 'user-input' to the next line and leaving them all 4-space-indented would be better.

Hwesta:
It's promising, but I have a few Python best practices/style suggestions.

A few general Python tips: 
* Scripts shouldn't have anything at the top level - everything should be in an function or the `if '__name__' == __main__` which only runs when it's a run as a script (not on import).
* Classes are only declared at the top level.
* Python prefers snake_case for variable & function names (CamelCase for classes).

So, some suggestions for fixing this:
* In models.init, put the class declaration and `Base = declarative_base()` back at the top level. Having an init function is a good idea (like Django has setup()), but it should only contain thing the stuff that needs config.
* In transfers, add a setup function that calls the `models.init`, and configures logging.  Also, make the config file var available globally. This is useful for the next point
* Add a `get_setting` that parses the config file and returns the result, similar to what Archivematica has.
* Since pidFileName/pidFileDirectory, logDirectory/logFileName and databaseDirectory/databaseName are only ever used together, why don't we combine them (pidfile, logfile and database?).

Some suggestions for what this might look like

```python
def get_setting(setting, default=None):
    config = configparser.SafeConfigParser()
    config.read(CONFIG_FILE)
    config.get(setting, default)
    
# Example:
pid_file = get_setting('pidfile')
```

```python
# At the top of the file
CONFIG_FILE = None

# Later...
def setup(config_file):
    global CONFIG_FILE
    CONFIG_FILE = config_file
    init(get_setting('database'))  # This is models.init

    # Configure logging
    default_logfile = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'automate-transfer.log')
    CONFIG = {
        'version': 1,
        'disable_existing_loggers': False,
        'formatters': {
            'default': {
                'format': '%(levelname)-8s  %(asctime)s  %(filename)s:%(lineno)-4s %(message)s',
                'datefmt': '%Y-%m-%d %H:%M:%S',
            },
        },
        'handlers': {
            'console': {
                'class': 'logging.StreamHandler',
                'formatter': 'default',
            },
            'file': {
                'class': 'logging.handlers.RotatingFileHandler',
                'formatter': 'default',
                'filename': get_setting('logfile', default_logfile),
                'backupCount': 2,
                'maxBytes': 10 * 1024,
            },
        },
        'loggers': {
            'transfer': {
                'level': 'INFO',  # One of INFO, DEBUG, WARNING, ERROR, CRITICAL
                'handlers': ['console', 'file'],
            },
        },
    }
```

```python
    # Other parsing code here...
    parser.add_argument('-c', '--config-file', help='Configuration file (log/db/PID files)', default='/etc/archivematica/automation-tools/transfers.conf')
    args = parser.parse_args()

    setup(args.config_file)
    # Rest of function...
```

Let me know if you need any clarifications about what I meant or what would be better.

hakamine:
Thank you for all the advice and suggestions. Made changes following those, code looks more readable now (and is hopefully ready for merging)

transfers/transfer.py
```diff
@@ -158,8 +178,8 @@ def get_accession_id(dirname):
     script_path = os.path.join(THIS_DIR, 'get-accession-number')
     try:
         p = subprocess.Popen([script_path, dirname], stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    except FileNotFoundError:
-        LOGGER.info('%s does not exist.', script_path)
+    except:
```
Hwesta: Should be `except Exception:` so it doesn't catch `KeyboardInterrupt` etc https://docs.python.org/3/library/exceptions.html#exception-hierarchy Any exception handling should specify what exceptions we're expecting to catch, or at a minimum `Exception`.

transfers/transfer.py
```diff
@@ -301,16 +321,16 @@ def start_transfer(ss_url, ts_location_uuid, ts_path, depth, am_url, user_name,
...
 
     # Run all scripts in pre-transfer directory
     # TODO what inputs do we want?
     run_scripts('pre-transfer',
-        resp_json['path'],  # Absolute path
-        'standard',  # Transfer type
-    )
+                resp_json['path'],  # Absolute path
+                'standard',  # Transfer type
+                )
```
Hwesta: Nitpick about moving `'pre-transfer'` to the next line instead of re-indenting everything applies here too.

transfers/transfer.py
```diff
@@ -41,41 +44,58 @@ def fsdecode(filename):
...
+
+CONFIG_FILE = None
+
+
+def get_setting(setting, default=None):
+    config = configparser.SafeConfigParser()
+    try:
+        config.read(CONFIG_FILE)
+        return config.get('transfers', setting)
+    except:
```
Hwesta: `except Exception:` here too.

transfers/transfer.py
```diff
@@ -41,41 +44,58 @@ def fsdecode(filename):
...
+        return default
+
+
+def setup(config_file):
+    global CONFIG_FILE
+    CONFIG_FILE = config_file
+    models.init(get_setting('databasefile', os.path.join(os.path.dirname(__file__), 'transfers.db')))
+
+    # Configure logging
+    default_logfile = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'automate-transfer.log')
```
Hwesta: This and default database file could use `THIS_DIR` instead of `os.path.abspath(os.path.dirname(__file__))`

transfers/transfer.py
```diff
@@ -11,12 +11,15 @@
 import base64
 import logging
 import logging.config  # Has to be imported separately
+import models
```
Hwesta: I usually separate project-specific import from Python builtin imports, which is why it was further down (sorry, I didn't mention that earlier).  Also, py2/3 compatibility needs this to be `from . import models`

Hwesta:
Looks much better now, thanks!  A couple of minor comments again, and you can probably squash the refactor commit into the initial one if you want.

transfers/transfer.py
hakamine: When replacing `import models` with `from . import models` getting the following error:
`ValueError: Attempted relative import in non-package`
I think it is related to http://stackoverflow.com/questions/11536764/attempted-relative-import-in-non-package-even-with-init-py
As we are running transfers.py as a standalone script, would it be better stay with `import models` ? Or does it break anything in python3 (I am testing with python 2.7)

Hwesta: In python3 the import will fail, because imports are absolute by default and there's nothing called 'models' in the global packages.  Does adding `absolute_import` to the `from __future__ import` line fix it?

hakamine: Unfortunately using `absolute_import` didn't help.
I just tested with a virtualenv created with python 3 (3.4.3), and got an error `SystemError: Parent module '' not loaded, cannot perform relative import` when using `from . import models` However `import models` seem to work fine. So for me `import models` is working for python 2 and 3, when calling the script as the examples in the README. 
Do you run the tools in a different way?

Hwesta: Hm. I've been running `py.test` from the same directory with a python 2.7 or a python 3.5 virtualenv.  For me, if I change it to `import models` I get the error ` ImportError: No module named 'models'` in py3 and it works in py2.  Does it change anything if you move it below the sys.path modification?

hakamine:
As mentioned in http://stackoverflow.com/questions/11536764/attempted-relative-import-in-non-package-even-with-init-py there is an issue invoking python scripts when relative imports are used (`from . import models`). Using the `-m` flag in python is a workaround that seems to fix this problem. README has been updated with this information (so hopefully it's ready to merge this time)

etc/transfer-script.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# transfer script example
+# /etc/archivematica/automation-tools/transfer-script.sh
+cd /usr/lib/archivematica/automation-tools/transfers/
+/usr/share/python/automation-tools/bin/python transfer.py --user <user>  --api-key <apikey> --transfer-source <transfer_source_uuid>
```
Hwesta: Update this to use `-m transfers.transfer` ?

Hwesta:
Looks great!  Good find with the `-m`.  :+1:

etc/transfer-script.sh
hakamine: sorry! fixed

PR #20, opened 2016-05-06
transfers: archivesspace_ids.py path fix

hakamine:
In bags, metadata directory is in data directory, refs #9728

### DIFF
--- a/transfers/examples/pre-transfer/archivesspace_ids.py
+++ b/transfers/examples/pre-transfer/archivesspace_ids.py
@@ -7,11 +7,16 @@ import errno
 import os
 import sys
 
+
 def main(transfer_path):
     """
     Generate archivesspaceids.csv with reference IDs based on filenames.
     """
-    archivesspaceids_path = os.path.join(transfer_path, 'metadata', 'archivesspaceids.csv')
+    data_path = transfer_path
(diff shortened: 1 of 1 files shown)


Hwesta:
Do we want this to only work for bags then, but not other types of transfers?  Maybe we could add to the beginning
```python
data_path = transfer_path
if os.path.exists(os.path.join(transfer_path, 'data')):  # Assume this is a bag
    data_path = os.path.join(transfer_path, 'data')
```
and use `data_path` for the places that you modified?

hakamine:
Could it be possible to have a more broad solution that also works for standard transfers? (I wasn't sure about how to do in that case, could there be any script in the existing codebase that I could refer to? )

Hwesta:
I think the solution I proposed will also work for standard transfers.  Most incoming transfers have the format:
```
logs/
metadata/
objects/
```
or just
```
metadata/
<objects in the root directory>
```
Bags have a more specific structure:
```
<misc bag file>
data/
```

Archivematica looks for a `metadata` directory at the root level when it looks for metadata files, and bags are restructured to put the contents of the data directory in the root of the unit. So by writing to a file in the `metadata` directory (or the `data/metadata` directory if it exists for a bag), it should work for both bags and non-bags.

hakamine:
Thank you for the explanation. Fixed as proposed.

Hwesta:
:+1:

PR #22, opened 2016-06-28
Dev/issue 9910 storage

mcantelon:
Script for automating moving of packages from one location to another:

* Get first UPLOADED or MOVING status package in source location (using REST call to SS), ordering results by UUID (to keep query results consistant)
* If package status is MOVING, do nothing, otherwise initiate move


Hwesta:
Please update the README to describe the new tool.

common/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+    logging.config.dictConfig(CONFIG)
+
+
+def open_pid_file(pid_file, logger=None):
+    try:
+        # Open PID file only if it doesn't exist for read/write
+        f = os.fdopen(os.open(pid_file, os.O_CREAT | os.O_EXCL | os.O_RDWR), 'r+')
+    except OSError,e:
+         if logger:
```
Hwesta: It makes more sense to assume the logger exists as a module level variable, rather than passed in. Let's define the logger at the top of the file and not pass it as a parameter to the helper functions.

common/utils.py
```diff
@@ -0,0 +1,91 @@
...
+            'storage': {
+                'level': 'INFO',  # One of INFO, DEBUG, WARNING, ERROR, CRITICAL
+                'handlers': ['console', 'file'],
+            },
+        },
+    }
+    logging.config.dictConfig(CONFIG)
+
+
+def open_pid_file(pid_file, logger=None):
```
Hwesta: Docstring please

common/utils.py
```diff
@@ -0,0 +1,91 @@
...
+            'file': {
+                'class': 'logging.handlers.RotatingFileHandler',
+                'formatter': 'default',
+                'filename': filename,
+                'backupCount': 2,
+                'maxBytes': 10 * 1024,
+            },
+        },
+        'loggers': {
+            'storage': {
```
Hwesta: Since this is a common util now, it should let the caller define what logger to listen for.  And since we're adding a common logger, it should be configured here by default too.

common/utils.py
```diff
@@ -0,0 +1,91 @@
...
+def get_setting(config_file, config_name, setting, default=None):
+    config = configparser.SafeConfigParser()
+    try:
+        config.read(config_file)
+        return config.get(config_name, setting)
+    except Exception:
+        return default
+
+
+def configure_logging(filename): #get_setting('logfile', default_logfile)
```
Hwesta: Docstring please. Also, is the comment still needed?

common/utils.py
```diff
@@ -0,0 +1,91 @@
...
+        },
+    }
+    logging.config.dictConfig(CONFIG)
+
+
+def open_pid_file(pid_file, logger=None):
+    try:
+        # Open PID file only if it doesn't exist for read/write
+        f = os.fdopen(os.open(pid_file, os.O_CREAT | os.O_EXCL | os.O_RDWR), 'r+')
+    except OSError,e:
```
Hwesta: This syntax for exception handling is deprecated in Python 3 - it should be `except OSError as e`.  Also, it's not necessary if you use `exc_info=True` in the logging statement below.

storage/move_packages.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python
+"""
+Automate Package Moving
+
+Helper script to automate moving packages in the Archivematica Storage Service.
```
Hwesta: :100:

storage/move_packages.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Configure logging
+    default_logfile = os.path.join(THIS_DIR, 'automate-storage.log')
+    logfile = get_setting('logfile', default_logfile)
+    utils.configure_logging(logfile)
+
+
+def get_first_eligible_package_in_location(ss_url, location_uuid):
+    """
+    Get first package in a location that has a status of either UPLOADED or MOVING.
```
Hwesta: This should probably also check for VERIFIED, but it should be discussed further.

storage/move_packages.py
```diff
@@ -0,0 +1,138 @@
...
+
+    :param str ss_url: Storage service URL
+    :param str package_uuid: UUID of package to move
+    :param str location_uuid: UUID of location to move package to
+    :returns: Dict representing JSON response.
+    """
+    LOGGER.info("Moving package %s to location %s", package_uuid, location_uuid)
+
+    post_url = '%s/api/v2/file/%s/move/' % (ss_url, package_uuid)
+    post_body = json.dumps({'location_uuid': location_uuid})
```
Hwesta: requests is actually smart enough to encode JSON itself, so you don't need to stringify it. http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests

storage/move_packages.py
```diff
@@ -0,0 +1,138 @@
...
+
+    r = requests.post(post_url,
+                      data=post_body,
+                      headers={'content-type': 'application/json'})
+    LOGGER.debug('Response: %s', r)
+    LOGGER.debug('Response text: %s', r.text)
+    if r.status_code != 200:
+        return None
+
+    return json.loads(r.text)
```
Hwesta: requests provides a convenience function `r.json()` to do this.

storage/move_packages.py
```diff
@@ -0,0 +1,138 @@
...
+    if utils.open_pid_file(pid_file, LOGGER) is None:
+        return 0
+
+    # Check statuis of last package and attempt move
+    move_result = None
+    package = get_first_eligible_package_in_location(ss_url, from_location_uuid)
+    if package is None:
+        LOGGER.info('No packages remain in location, nothing to do.')
+    elif package['status'] == 'MOVING':
+        LOGGER.info('Current package still processing, nothing to do.')
```
Hwesta: Might be helpful to log what the current package (UUID?) is.

storage/move_packages.py
```diff
@@ -0,0 +1,138 @@
...
+    else:
+        LOGGER.info('Moving package.')
```
Hwesta: Also might be useful to log which package is being moved here.

Hwesta:
Looks pretty good!  Can you add tests, and verify it works in Python3 as well? I noticed some python2-only syntax.

hakamine:
It looks like the code in this branch needs the addition of SS authentication in order to work with SS branch dev/issue-9909-move-endpoint-0_10_0  ( 0.9.x + move endpoint code )

PR #23, opened 2016-07-13
Transfers: Specify verbosity level

Hwesta:
Add -v and -q paramaters to move the logging level up or down.


sevein:
Can we have an extra `--log-level=INFO|DEBUG|WARNING|ERROR|CRITICAL`?

Hwesta:
Instead of or in addition to -q/-v? How do you suggest having --log-level interact with -q/-v?

sevein:
If `--log-level` is used, I think it should override whatever is the result of combining `-q` and `-v`.

Hwesta:
Sure, that seems reasonable.

Hwesta:
Updated - how does this look? I'll probably squash the commits before merge.

sevein:
👍

PR #24, opened 2016-08-30
transfers: add_metadata hook writes csv

Hwesta:
add_metadata example hook writes a metadata.csv instead of a metadata.json. Not all paths correctly handle metadata.json, so this will reduce problems with other types.


sevein:
👍

hakamine:
👍 tested and working for unzipped bags

PR #17, opened 2016-04-19
Add example automated reingest script & docs

Hwesta:
Add a script showing how the Archivematica & storage service API can be called to start and approve reingest. Split the docs into an automated transfers & automated reingest section.

Relates to artefactual/archivematica#391 and artefactual/archivematica-storage-service#100

PR #25, opened 2016-10-19
Added amclient module: can hide approved transfers.

jrwdunham:


transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+        if b64decode:
+            return self.b64decode_ts_location_browse(result)
+        return result
+
+    def b64decode_ts_location_browse(self, result):
+        """Base64-decode the results of a call to SS GET
+        /location/UUID/browse/.
+        """
+        def dec(thing):
+            return base64.b64decode(getattr(thing, 'encode')('utf8'))
```
Hwesta: Why not `base64.b64decode(thing.encode('utf8'))` ?

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+            result['properties'] = dict((dec(k), v) for k, v in
+                                        result['properties'].items())
+            return result
+        except Exception:
+            return result
+
+    def search_package(self, params):
+        """SS GET /search/package?<GET_PARAMS>."""
+        return self._call_url_json(
+            '{}/api/v2/search/package/'.format(self.ss_url), params)
```
Hwesta: Resources in the storage service already allow you to filter by certain fields. Can this be done using those instead of requiring the new /search/ endpoint? eg /api/v2/file/?package_type=AIP

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+            r = self.search_package({'package_type': ptype})
+        packages = packages + r['results']
+        if r['next']:
+            packages = self.get_all_packages(ptype, packages, r['next'])
+        return packages
+
+    def aip2dips(self):
+        """Get all DIPS created from AIP with UUID ``self.aip_uuid``."""
+        dips = self.get_all_dips()
+        return [d for d in dips if
```
Hwesta: Can this be done by only getting DIPs that related to an AIP UUID using tastypie's filters? https://github.com/artefactual/archivematica-storage-service/blob/stable/0.9.x/storage_service/locations/api/resources.py#L434 should be relevant.

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+                      [d['uuid'] for d in dips if
+                       d['current_path'].endswith(a['uuid'])])
+                     for a in aips])
+
+    def download_dip(self):
+        """Download the DIP with UUID ``self.dip_uuid``."""
+        url = '{}/api/v2/file/{}/download/'.format(self.ss_url, self.dip_uuid)
+        r = requests.get(
+            url,
+            params={'username': self.ss_user_name,
```
Hwesta: Should this use `_ss_auth`?

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+            with open(local_filename, 'wb') as f:
+                for chunk in r.iter_content(chunk_size=1024):
+                    if chunk:
+                        f.write(chunk)
+            print(local_filename)
+            return local_filename
+        else:
+            print('Unable to download DIP {}'.format(self.dip_uuid))
+
+    def reingest(self):
```
Hwesta: These functions look like a duplicate of #17?

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+        r = requests.post(url, json=payload, headers=auth_header)
+        try:
+            assert r.status_code == 202
+            print('Reingest of AIP {} initiated.'.format(self.aip_uuid))
+        except AssertionError:
+            print('ERROR: something went wrong when attempting to reingest AIP'
+                  ' {}.'.format(self.aip_uuid))
+        self.approve_reingest(self.aip_uuid)
+
+    approve_reingest_retry_count = 5
```
Hwesta: This constant should be inside the function that uses it, or at the top of the class with the rest of the constants. Since it's only used once I'm inclined to say the former.

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+
+    def _approve_reingest(self, reingest_uuid):
+        """Approve re-ingest with UUID ``reingest_uuid``.
+
+        :returns: UUID of the approved ingest or None.
+        """
+        time.sleep(5)
+        # List available ingests
+        get_url = '{}/api/ingest/waiting'.format(self.am_url)
+        params = {'username': self.am_user_name, 'api_key': self.am_api_key}
```
Hwesta: Should this use `_am_auth`?

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+        waiting_transfers = requests.get(get_url, params).json()
```
Hwesta: Should this use `_call_url_json`?

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+        encoding = sys.getfilesystemencoding()
+        if isinstance(filename, str):
+            return filename
+        elif isinstance(filename, unicode):
+            return filename.encode(encoding)
+        else:
+            raise TypeError("expect bytes or str, not %s" %
+                            type(filename).__name__)
+
+    def fsdecode(filename):
```
jrwdunham: Ok. Removed.

transfers/amclient.py
```diff
@@ -0,0 +1,518 @@
...
+
+    def _ss_auth(self):
+        return {
+            'username': self.ss_user_name,
+            'api_key': self.ss_api_key
+        }
+
+    def hide_transfer(self, transfer_uuid):
+        """GET transfer/<UUID>/delete/."""
+        return requests.delete(
```
jrwdunham: Ok. I've modified call_url_json to a func (no need to be a method) and it takes a method string param and calls requests.request instead of requests.get and it logs the method.

transfers/amclient.py
```diff
@@ -0,0 +1,487 @@
+#!/usr/bin/env python
+"""Archivematica Client.
+
+Module and CLI that holds functionality for interacting with the various
+Archivematica APIs.
+
+KNOWN BUG: if you call ./amclient.py with no args/options, you'll get an error.
+See https://github.com/aliles/begins/pull/60 for the PR that fixes this on Py3.
+"""
+from __future__ import print_function, unicode_literals
```
Hwesta: :100:

transfers/amclient.py
```diff
@@ -0,0 +1,487 @@
...
+def _call_url_json(url, params, method='GET'):
+    """Helper to GET a URL where the expected response is 200 with JSON.
+    :param str url: URL to call
+    :param dict params: Params to pass to requests.get
+    :returns: Dict of the returned JSON or None
+    """
+    if method.upper() in ('GET', 'POST', 'DELETE', 'PUT', 'HEAD', 'OPTIONS'):
+        method = method.upper()
+    else:
+        method = 'GET'
```
Hwesta: Could this if/else be `method = method.upper()`?  That would allow eg PATCH, requests already defaults to GET, and so is the default parameter of this function.

transfers/amclient.py
```diff
@@ -0,0 +1,487 @@
...
+        except (binascii.Error, TypeError):
+            LOGGER.warning('Failed to base64-decode file or directory names in'
+                           ' output from GET call to SS'
+                           ' /location/UUID/browse/: %s', result)
+        return thing
+    try:
+        result['directories'] = [dec(d) for d in result['directories']]
+        result['entries'] = [dec(e) for e in result['entries']]
+        result['properties'] = dict((dec(k), v) for k, v in
+                                    result['properties'].items())
```
Hwesta: Dictionary comprehensions are pretty great. This can be `{(dec(k): v for k, v in result['properties'].items()}`

tests/test_amclient.py
```diff
@@ -0,0 +1,344 @@
...
+SS_USER_NAME = 'test'
+SS_API_KEY = '5de62f6f4817f903dcfac47fa5cffd44685a2cf2'
+TMP_DIR = '.tmp-dip-downloads'
+TRANSFER_SOURCE_UUID='7609101e-15b2-4f4f-a19d-7b23673ac93b'
+UUID_PATT = re.compile(
+    '^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$')
+
+
+def is_uuid(thing):
+    return UUID_PATT.search(thing) is not None
```
Hwesta: This is a duplicate of what's in amclient - should probably just use that one.

transfers/amclient.py
Hwesta: Looks good!

transfers/amclient.py
jrwdunham: Sounds good.

tests/test_amclient.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+class TmpDir:
+    """Context manager to clear and create ``TMP_DIR`` and destroy it after
+    usage.
+    """
+
+    def __enter__(self):
```
jrwdunham: ok.

transfers/amclient.py
jrwdunham: Hereby consider me officially a dict comp convert.

jrwdunham:
Ok. I think I've addressed this round of concerns. I've changed the CLI to use argparse again. I got subcommands working. What do you think now?

Do you think tests targeting the command line interface are necessary or overkill?

Hwesta:
Nice job on the subcommands! 

> Do you think tests targeting the command line interface are necessary or overkill?

Probably overkill.

In python3, json.dumps() fails when the dict contains bytestrings (eg. in the command `transferables`).  "Works" in python2 until it encounters non-ASCII.  The results should be encoded (utf8?) before printing.  Interestingly this works for `unapproved-transfers` even when unicode provided, and none of the other commands seem to have this problem because the filenames are sanitized.

jrwdunham:
Nice catch about the json.dumps fail. I believe that simply modifying the return of ``b64decode_ts_location_browse`` to ``return thing.decode('utf8')`` in solves the issue. The ``return response.json()`` in ``_call_url_json`` should return Unicode (<str>) in all other cases.

Hwesta:
Looks good!  Breaks on non-unicode (& non ASCII) encodings, but that's not really a concern right now.

PR #26, opened 2016-12-07
Add split transfers script

Hwesta:
Add split_transfers script, previously at https://gist.github.com/sevein/617c12566ed3580ee566

This splits up transfers before automate transfers runs so that very large transfers can be processed automatically on smaller Archivematica pipelines.  It also splits up the contents of a `metadata.csv` file across directories.

This adds a `--metadata-only` flag that will update the metadata.csv but not move files.


sevein:
Nice clean-ups! :+1:

PR #29, opened 2016-12-22
transfer.py : fix approval API call bug

timothyryanwalsh:
Trailing slash in approve transfer post was resulting in 404 response.

### DIFF
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -382,7 +382,7 @@ def approve_transfer(directory_name, url, am_api_key, am_user):
         LOGGER.debug("Found waiting transfer: %s", a['directory'])
         if fsencode(a['directory']) == directory_name:
             # Post to approve transfer
-            post_url = url + "/api/transfer/approve/"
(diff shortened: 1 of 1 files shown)


Hwesta:
Good catch! This is caused by a bug in declaring URLs in Archivematica, which is fixed by artefactual/archivematica#551 and will be in the upcoming AM 1.6 release. Your change should work with both versions though.

timothyryanwalsh:
Glad it's gotten resolved! If I understand you correctly, this means when we upgrade to 1.6, the trailing slash in "/api/transfer/approve/" should work fine?

Hwesta:
Correct, once you upgrade to AM 1.6 "/api/transfer/approve/" should work.

Hwesta:
This bug is fixed in AM 1.6, and the fix was backported to stable/1.5.x where it was introduced.  I don't think this PR is needed anymore, since the problem was fixed elsewhere.  Thanks for the bug report and proposed fix!

PR #30, opened 2017-01-02
Dev/issue 9340 hide completed ingests

jrwdunham:


tests/test_amclient.py
```diff
@@ -17,6 +17,7 @@
 SS_URL = 'http://192.168.168.192:8000'
 AM_USER_NAME = 'test'
 AM_API_KEY = '3c23b0361887ace72b9d42963d9acbdf06644673'
+AM_API_KEY_2 = 'ff8fa7f316168508bf9bc743ac589ec55bfd611a'
```
Hwesta: Instead of adding a second testing key, you can modify the recorded cassette to use the existing one.  

Alternately, we could configure vcr.py not to record the api key, to avoid problems like this, and update the cassettes to remove the api_key param.  eg https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-information-from-http-querystring

transfers/amclient.py
```diff
@@ -382,46 +394,77 @@ def hide_transfer(self, transfer_uuid):
             method='DELETE'
         )
 
+    def hide_unit(self, unit_uuid, unit_type):
```
Hwesta: Is hide_transfer still needed? If so, can it be converted into a wrapper for hide_unit?

jrwdunham: Nope. not needed, just forgot to delete it. Thanks.

hakamine:
Confirmed to be working in a client's testing server👍

PR #31, opened 2017-01-11
FIX creation of database

remileduc:
First time you run the automation-tools, the database file need to be created.

Currently, you'll have an error saying that it is impossible to open the file.

This fix will create the file if it doesn't exist. In case it is not possible to create it (permission), the standard error (access denied) will be shown.

### DIFF
--- a/transfers/models.py
+++ b/transfers/models.py
@@ -1,3 +1,5 @@
+from os.path import isfile
+
 from sqlalchemy import create_engine
 from sqlalchemy import Sequence
 from sqlalchemy import Column, Binary, Boolean, Integer, String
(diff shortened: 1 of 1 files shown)


transfers/models.py
```diff
@@ -22,6 +24,10 @@ def __repr__(self):
 
 
 def init(databasefile):
+    if not isfile(databasefile):
+        # We create the file
+        with open(databasefile, "a"):
+            pass
```
remileduc: Here, the other solution is just to put
```python
with open(databasefile, "a"):
    pass
```
(so remove the condition)

But the disadvantage is that it will open the file everytime (but the code is simpler).

There is also the solution to use this function: https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch
Unfortunately it doesn't seem to exist in python 2.

PR #33, opened 2017-07-11
transfers/amclient: config_file should be log_file

hakamine:


jrwdunham:
👍 Yes, this looks right. I think that was a typo. 

hakamine, I think (relatedly) you could also remove the `CONFIG_FILE` constant and the `get_setting` function since those don't appear to be used. But if you don't want to do that in this PR that's fine with me too.

sevein:
@jrwdunham @hakamine - I've just made that last change. Should we merge?

hakamine:
@jdunham @sevein thank you for the CR/fixes, I'll rebase and merge

PR #41, opened 2017-11-30
Create dips.atom_upload module and scripts

jraddaoui:
Sends the DIP to the AtoM host using rsync and executes a deposit request to the
AtoM instance. A passwordless SSH connection is required to the AtoM host for the
user running this script and it must be already added to the list of known hosts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ The Automation Tools project is a set of python scripts, that are designed to au
   - [Configuration](#configuration-1)
     - [Parameters](#parameters-1)
     - [Getting Storage Service API key](#getting-storage-service-api-key)
+- [DIP upload to AtoM](#dip-upload-to-atom)
+  - [Configuration](#configuration-2)
+    - [Parameters](#parameters-2)
 - [Related Projects](#related-projects)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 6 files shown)


jraddaoui:
Hi @sevein, I was thinking on adding unit tests for this script but it only executes a `rsync` command, which may fail depending on the SSH configuration or on who is running the script, and then posts a deposit request to an AtoM instance without doing a lot with the response after. Considering that we're storing the request responses the first time the test is executed, I can't see the point. But probably there is one, so I'd appreciate any advice.

sevein:
If you wanted to test `main` from an unit test I think first you look at the dependencies, e.g. `rsync` and `deposit` functions. You want to isolate the behaviour of `main` so you replace the dependencies by mocks that simulate the behaviour of the real functions.

I've uploaded a gist to show you an example: https://gist.github.com/sevein/f7ca759cce11a29d3b816ba14dbefbee.

If `mock` is not available add it to the requirements. It's just a port of Python3's `unittest.mock` for Python2.

sevein:
Similarly, if you wanted to unit test `rsync` you can mock `subprocess.check_output`. You want to write at least a couple of tempts, one of them where `side_effect` is to return an exception but another one where it doesn't. It can be painful but I recommend spending some time reading https://docs.python.org/3/library/unittest.mock.html#module-unittest.mock - there is stuff there that you're going to need sooner or later so it's better to have a good mental map of what's possible.

jraddaoui:
Awesome, thanks @sevein!

PR #42, opened 2017-12-29
Add DIP creation job module and script

jraddaoui:
Creates a process than can be automated to create DIPs from all the AIPs
in an specified Storage Service location.

Allows the use of more parameters in the `amclient` methods to obtain
packages to be able to filter the results.


jrwdunham:
👍 lgtm

PR #43, opened 2018-01-08
Improvement: add "whitelist" for script file extensions

lower29:
The `run_scripts` function in `transfer.py` is currently very trusting in terms of what it should execute. It uses the execute bit in the file mode to determine if something is executable, and assumes it should be executed if this is set. Aside from the obvious security concern about executing user-contributed code, it also falls foul of the execution bit being set incorrectly, e.g. by some well-meaning recursive `chmod 777` etc. In particular, we have found that XML files with the executable bit set are considered for execution, which consequently fails.

This pull request improves the way that `run_scripts` works by adding a `script_extensions` whitelist that files should have before being considered executable by the code. This is configurable using the `scriptextensions` parameter in the config file. By default it is set to an empty list, meaning that all executable files are executed (retaining the current behaviour), but setting it to `.py:.sh` will cause any files that don't have a `.py` or `.sh` extension to be ignored and not considered for execution. This satisfies our use case, causing XML files that are "executable" to be ignored.

In addition, I've added some error handling to the `start_transfer` function. Previously, any error encountered in the call to `run_scripts` when processing files in the `pre-transfer` folder was not handled, and caused the entire `transfer.py` process to crash out, leaving its pid lock file in place and preventing any further executions to occur. Errors in the `run_scripts` call are now caught, and reported via the logger, before returning cleanly and allowing the pid lock file to be cleaned up before the process completes.

Lastly, this pull request includes a previous commit from @jhsimpson to make the `run_scripts` function look at the target of symlinks, using the `realpath` of the file rather than its source path. This was believed to be the reason why scripts were being considered for execution, but even if you follow the symlink then if the target file has its executable bit set then execution will still be attempted, hence the need for the `script_extensions` addition described above.

### DIFF
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -213,7 +213,7 @@ def run_scripts(directory, *args):
     LOGGER.debug('script_args: %s', script_args)
     for script in sorted(os.listdir(directory)):
         LOGGER.debug('Script: %s', script)
-        script_path = os.path.join(directory, script)
+        script_path = os.path.realpath(os.path.join(directory, script))
         if not os.path.isfile(script_path):
             LOGGER.info('%s is not a file, skipping', script)
             continue
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -211,14 +211,20 @@ def run_scripts(directory, *args):
         return
     script_args = list(args)
     LOGGER.debug('script_args: %s', script_args)
+    script_extensions = get_setting('scriptextensions', '').split(':')
+    LOGGER.debug('script_extensions: %s', script_extensions)
     for script in sorted(os.listdir(directory)):
         LOGGER.debug('Script: %s', script)
         script_path = os.path.realpath(os.path.join(directory, script))
         if not os.path.isfile(script_path):
-            LOGGER.info('%s is not a file, skipping', script)
+            LOGGER.info('%s is not a file, skipping', script_path)
             continue
         if not os.access(script_path, os.X_OK):
-            LOGGER.info('%s is not executable, skipping', script)
+            LOGGER.info('%s is not executable, skipping', script_path)
+            continue
+        script_name, script_ext = os.path.splitext(script)
+        if script_extensions and script_ext not in script_extensions:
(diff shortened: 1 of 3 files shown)


transfers/transfer.py
```diff
@@ -336,13 +342,19 @@ def start_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, ts_path, depth
...
+        run_scripts(
+            'pre-transfer',
+            resp_json['path'],  # Absolute path
+            'standard',  # Transfer type
+        )
+    except:
+        # An error occurred, log and recover cleanly, returning to main process
+        e = sys.exc_info()[0]
+        LOGGER.error("Failed to run pre-transfer scripts: %s", e)
+        return None
```
sevein: I think this block could be written as follows:

```python
try:
    ...
except Exception as err:
    LOGGER.error("Failed to run pre-transfer scripts: %s", err)
    return None
```

lower29: You're right, it could, but then it wouldn't [catch general exceptions](https://wiki.python.org/moin/HandlingExceptions), only those that extend from `Exception`, so I think this way is better (my choice was deliberate).

sevein: So why not to use `BaseException` instead? Bare excepts are generally not recommended. Bare excepts where used it in Py2 because you could have a custom exception not subclassing `Exception` but that's not possible anymore neither you have such kind of code.

sevein: https://docs.python.org/2/howto/doanddont.html#except

sevein: I've filed https://github.com/JiscRDSS/rdss-archivematica-automation-tools/issues/13 btw.

sevein: Feel free to ignore this. I think I am used to avoid bare excepts because my linter complains about it but it should not be a problem.

lower29: I've changed the `except` line to name `BareException`. I've also fixed the length of the summary line of the commit. If you're happy with these changes then please approve so we can merge.

I'm assuming by the fact you're nitpicking that you're okay with the main purpose of this commit, namely the use of the `script_extensions` whitelist and its config.

PR #45, opened 2018-02-15
Avoid auth. concatenation in next requests

jraddaoui:
We're sending the auth. parameters as part of the URL and they're
being returned in the next path used for the following request,
therefore, as we add them again in the next requests, they were
being sent multiple times, creating long request URLs.

The auth. should be sent in the header to avoid adding sensitive
data in the SS response but this commit avoids at least its
duplication.

The VCR cassetes have been updated to reflect the new behaviour.


mcantelon:
👍

PR #46, opened 2018-02-20
Refactors amclient.py

ross-spencer:
WIP. Hi @jrwdunham @jraddaoui can I request your review of this work? I wanted to separate the logic of the various components of amclient.py to make it clearer to extend in future. I want to base a script for a client of this work in the coming days. The summary is below.

Before this can be merged, I am having a problem with ModuleImportError in the unit tests. I have tried my best to solve this today but to no avail. The scripts work independently as executable scripts, but it seems there maybe an issue testing that at the package level, and it seems like this has changed in Python 3.6. The tests pass in 2.7. 

Your view on the refactor, and the test issue would be greatly appreciated. I'll try and take a look later.

* Creates new logging configuration script to reduce redundancy
* Separates responsibility of amclient, defaults, and cli args
* Cleans up argparse output using metavar
* Adds Conformity to PEP8 to amclient.py and transfers.py
* Returns user friendly string if arg evaluation fails in main
* Improved error return from JSON request function in amclient.py
* Adds .db and .lck files to .gitignore
* Works at module and executable level
* E402 flake8 warning ignored in Tox to enable module/exe to work
* Captures connection errors in transfers.py better
* Initial changes provides scope for future refactoring as required
* Addresses #47

### DIFF
--- a/tests/test_amclient.py
+++ b/tests/test_amclient.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 """To run the tests::
 
     $ python -m unittest tests.test_amclient
@@ -11,6 +10,7 @@ import unittest
 import vcr
 
 from transfers import amclient
+from transfers import errors
 
 
 AM_URL = 'http://192.168.168.192'
@@ -24,6 +24,7 @@ TRANSFER_SOURCE_UUID = '7609101e-15b2-4f4f-a19d-7b23673ac93b'
 
 
 class TmpDir:
+
     """Context manager to clear and create a temporary directory and destroy it
     after usage.
     """
@@ -118,7 +119,7 @@ class TestAMClient(unittest.TestCase):
         completed_transfers = amclient.AMClient(
             am_api_key='bad api key', am_user_name=AM_USER_NAME,
             am_url=AM_URL).completed_transfers()
-        assert completed_transfers is None
+        assert completed_transfers is errors.ERR_INVALID_RESPONSE
 
     @vcr.use_cassette(
         'fixtures/vcr_cassettes/unapproved_transfers_transfers.yaml')
--- a/transfers/amclient.py
+++ b/transfers/amclient.py
@@ -1,224 +1,44 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 """Archivematica Client.
 
 Module and CLI that holds functionality for interacting with the various
 Archivematica APIs.
 """
+
 from __future__ import print_function, unicode_literals
 
-import argparse
 import binascii
 import base64
-from collections import defaultdict, namedtuple
+from collections import defaultdict
 import json
-import logging
-import logging.config  # Has to be imported separately
 import os
 import pprint
 import re
 import sys
 
 import requests
-from six import binary_type, text_type
+import urllib3
+
+# AM Client module configuration
+
+# Allow execution as an executable and the script to be run at package level
+# by ensuring that it can see itself.
+sys.path.append('../')
+
+from transfers import loggingconfig
+from transfers import defaults
+from transfers import amclientargs
+from transfers import errors
 
 
-try:
-    from os import fsencode
-except ImportError:
-    def fsencode(filename):
-        """Cribbed & modified from Python3's OS module to support Python2."""
-        encoding = sys.getfilesystemencoding()
-        if isinstance(filename, binary_type):
-            return filename
-        elif isinstance(filename, text_type):
-            return filename.encode(encoding)
-        else:
-            raise TypeError("expect bytes or str, not %s" %
-                            type(filename).__name__)
+def get_logger(log_file_name, log_level):
+    return loggingconfig.setup(log_level, log_file_name, "amclient")
 
 
-THIS_DIR = os.path.abspath(os.path.dirname(__file__))
-DEFAULT_LOGFILE = os.path.join(THIS_DIR, 'amclient.log')
-LOGGER = logging.getLogger('amclient')
-RETRY_COUNT = 5
-DEF_AM_URL = 'http://127.0.0.1'
-DEF_SS_URL = 'http://127.0.0.1:8000'
-DEF_USER_NAME = 'test'
-UUID_PATT = re.compile(
-    '^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$')
-UNDECODABLE = 'UNABLE TO DECODE'
-UNDEC_MSG = ('Unable to decode a transfer source component; giving up and'
-             ' returning {0}'.format(UNDECODABLE))
-
-
-# Reusable argument constants (for CLI).
-Arg = namedtuple('Arg', ['name', 'help', 'type'])
-AIP_UUID = Arg(
-    name='aip_uuid',
-    help='UUID of the target AIP',
-    type=None)
-AM_API_KEY = Arg(
-    name='am_api_key',
-    help='Archivematica API key',
-    type=None)
-DIP_UUID = Arg(
-    name='dip_uuid',
-    help='UUID of the target DIP',
-    type=None)
-SS_API_KEY = Arg(
-    name='ss_api_key',
-    help='Storage Service API key',
-    type=None)
-TRANSFER_SOURCE = Arg(
-    name='transfer_source',
-    help='Transfer source UUID',
-    type=None)
-
-
-# Reusable option constants (for CLI).
-Opt = namedtuple('Opt', ['name', 'metavar', 'help', 'default', 'type'])
-AM_URL = Opt(
-    name='am-url',
-    metavar='URL',
-    help='Archivematica URL. Default: {0}'.format(DEF_AM_URL),
-    default=DEF_AM_URL,
-    type=None)
-AM_USER_NAME = Opt(
-    name='am-user-name',
-    metavar='USERNAME',
-    help='Archivematica username. Default: {0}'.format(DEF_USER_NAME),
-    default=DEF_USER_NAME,
-    type=None)
-DIRECTORY = Opt(
-    name='directory',
-    metavar='DIR',
-    help='Directory path to save the DIP in',
-    default=None,
-    type=None)
-OUTPUT_MODE = Opt(
-    name='output-mode',
-    metavar='MODE',
-    help='How to print output, JSON (default) or Python',
-    default='json',
-    type=None)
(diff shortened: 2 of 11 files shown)


jraddaoui:
Hi @ross-spencer, the imports issue can be fixed by using `from . import defaults` instead of `import defaults` and the same for other relative imports you added. This may be related:

https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code

jraddaoui:
Sorry, wrong button.

ross-spencer:
@jraddaoui @jrwdunham thanks for your advice. It was a pain to fix. Joel made another PR for me to look at and I used the change from that plus imported the module to the path which you'll see in the code as `sys.path.append('../')`. I was finding three or four different ways of importing these files as modules/scripts etc. Getting it to work standalone and then watching the tests fail, and vice versa. I think this is the best I've found to allow us to pull apart some of the code to make individual scripts a little simpler, and have everything work as expected. I had to disable a flake8 warning around imports, but I hope that's okay.

I'm going to have a look in more detail at our client's requirements and maybe use this base as a premise. Let me know what changes I can make as required. Hope it's mostly okay though.

transfers/transfer.py
```diff
@@ -224,29 +227,40 @@ def run_scripts(directory, *args):
...
-    :param set completed: Set of the paths of completed transfers. Ideally, relative to the same transfer source location, including the same path_prefix, and at the same depth.
-    :param bool see_files: Return files as well as folders to become transfers.
+    :param depth: Depth relative to path_prefix to create a transfer from.
+                  Should be 1 or greater.
+    :param set completed: Set of the paths of completed transfers. Ideally,
+                          relative to the same transfer source location,
+                          including the same path_prefix, and at the same
+                          depth.
+    :param bool see_files: Return files as well as folders to become
+                           transfers.
```
jrwdunham: While I personally appreciate respecting the 80 character line limit, most of AM's codebase does not respect it, so it's probably not practical to fix all of the >80 chars lines in here. If you are going to do embark on massive refactors, it would make it a lot easier to code review if you separate content-ful commits from purely stylistic ones. Then the reviewer can go through the stylistic commit(s) and just confirm that nothing has been changed in terms of semantics/logic.

transfers/transfer.py
```diff
@@ -424,9 +464,11 @@ def main(am_user, am_api_key, ss_user, ss_api_key, ts_uuid, ts_path, depth, am_u
     pid_file = get_setting('pidfile', default_pidfile)
     try:
         # Open PID file only if it doesn't exist for read/write
-        f = os.fdopen(os.open(pid_file, os.O_CREAT | os.O_EXCL | os.O_RDWR), 'r+')
+        f = os.fdopen(
+            os.open(pid_file, os.O_CREAT | os.O_EXCL | os.O_RDWR), 'r+')
     except:
-        LOGGER.info('This script is already running. To override this behaviour and start a new run, remove %s', pid_file)
+        LOGGER.info(('This script is already running. To override this '
+                     'behaviour and start a new run, remove %s'), pid_file)
```
jrwdunham: You don't need the extra set of parens here.

transfers/transfer.py
```diff
@@ -463,13 +507,16 @@ def main(am_user, am_api_key, ss_user, ss_api_key, ts_uuid, ts_path, depth, am_u
...
+            'Waiting on user input, running scripts in user-input directory.')
         # TODO What inputs do we want?
         microservice = status_info.get('microservice', '')
         run_scripts(
             'user-input',
             microservice,  # Current microservice name
-            str(microservice != current_unit.microservice),  # String True or False if this is the first time at this wait point
+            str(microservice != current_unit.microservice),
+            # String True or False if this is the first time at this wait
+            # point
```
jrwdunham: Nitpick: I think the descriptive comment should go above the thing it describes.

transfers/amclient.py
```diff
@@ -1,224 +1,45 @@
...
-                                          help=subcommand.help)
-        for arg in subcommand.args:
-            subparser.add_argument(
-                arg.name, help=arg.help, type=arg.type)
-        for opt in subcommand.opts:
-            subparser.add_argument(
-                '--' + opt.name, metavar=opt.metavar, help=opt.help,
-                default=opt.default, type=opt.type)
-    return parser
+import urllib3
```
jrwdunham: I think that you need to add urllib3 to the requirements/base.txt file. urllib3 became a dependency of requests but only as of a certain version (I don't know exactly which). Because our requests requirement is vague, it's possible to have requests installed but not urllib3. The other option would be to pin down a more specific (higher) requests version in our requirements.

transfers/transfer.py
```diff
@@ -491,41 +541,76 @@ def main(am_user, am_api_key, ss_user, ss_api_key, ts_uuid, ts_path, depth, am_u
...
-    parser.add_argument('--quiet', '-q', action='count', default=0, help='Decrease the debugging output')
-    parser.add_argument('--log-level', choices=['ERROR', 'WARNING', 'INFO', 'DEBUG'], default=None, help='Set the debugging output level. This will override -q and -v')
+    parser.add_argument('--verbose', '-v', action='count',
+                        default=0, help='Increase the debugging output.')
+    parser.add_argument('--quiet', '-q', action='count',
+                        default=0, help='Decrease the debugging output')
+    parser.add_argument(
+        '--log-level', choices=['ERROR', 'WARNING', 'INFO', 'DEBUG'],
+        default=None, help='Set the debugging output level. This will \
+        override -q and -v')
```
jrwdunham: MegaNitpick: "The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation." From https://www.python.org/dev/peps/pep-0008/.

In the case of breaking strings across lines, I think the parenthesis with implicit concatenation strategy is better because you don't end up having strange sequences of whitespace in the middle of your strings. In this context it doesn't matter because argparse is getting rid of them for you, but in general, it's one reason beyond an appeal to authority for this preference.

transfers/transfer.py
```diff
@@ -491,41 +541,76 @@ def main(am_user, am_api_key, ss_user, ss_api_key, ts_uuid, ts_path, depth, am_u
...
-        level = args.quiet - args.verbose
-        level = max(level, -1)  # No smaller than -1
-        level = min(level, 2)  # No larger than 2
-        log_level = log_levels[level]
-    else:
-        log_level = args.log_level
+    log_level = loggingconfig.set_log_level(
+        args.log_level, args.quiet, args.verbose)
+
+    global CONFIG_FILE
```
jrwdunham: I think Python's `global` keyword should be avoided in general. I would like to avoid adding more `global`s to the codebase. Why not just add a `config_file` param to `get_setting`?

transfers/amclient.py
```diff
@@ -583,14 +375,37 @@ def download_aip(self):
...
+    args = argparser.parse_args()
     am_client = AMClient(**vars(args))
+
+    # Re-configure global LOGGER based on user provided parameters.
+    global LOGGER
+    LOGGER = get_logger(args.log_file, args.log_level)
+
     try:
-        getattr(am_client, 'print_{0}'.format(args.subcommand.replace('-', '_')))
+        func = getattr(am_client, '{0}'.format(
```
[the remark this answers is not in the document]
jrwdunham: If you do have a good reason to remove the `'print_`, then you should remove the call to `.format` since it would be unnecessary.

transfers/transfer.py
ross-spencer: I've given it another shot to make this section look a bit prettier and still respect the line limit (this is a tough script to make it adhere to that!). I hadn't anticipated that style/contents should be separated but will consider this in future.

transfers/transfer.py
ross-spencer: I do agree. I should have fixed that now with the next commit when it arrives.

transfers/transfer.py
ross-spencer: Should be good now.

transfers/transfer.py
ross-spencer: :+1:

transfers/amclient.py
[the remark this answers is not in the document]
ross-spencer: The primary reason was that in being verbose here we have an opportunity to write a user friendly output when there is an error condition, e.g. connecting on the wrong port:

```
ross-spencer@artefactual:~/git/artefactual/automation-tools/transfers:$ python amclient.py completed-transfers test
{u'message': u'Fetched completed transfers successfully.', u'results': []}
ross-spencer@artefactual:~/git/artefactual/automation-tools/transfers:$ python amclient.py completed-transfers test
Error connecting to the server, check amclient log
```

Where before, relying on the previous call and error handling we'd get a result of `Null` at the terminal, or, given a string returned from the calling function, would still have its formatting in place, e.g. as a return, `"Error connecting to the server, check amclient log\n"`

It does implement the expected behaviour which I think might be enough to keep this change as-is, but what do you think?

transfers/amclient.py
```diff
@@ -1,224 +1,45 @@
...
 
 Module and CLI that holds functionality for interacting with the various
 Archivematica APIs.
 """
+
 from __future__ import print_function, unicode_literals
 
-import argparse
+import os
+
```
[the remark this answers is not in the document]
ross-spencer: I hadn't appreciated this, but works for me.

transfers/amclient.py
```diff
@@ -583,14 +375,37 @@ def download_aip(self):
...
+    global LOGGER
+    LOGGER = get_logger(args.log_file, args.log_level)
+
     try:
-        getattr(am_client, 'print_{0}'.format(args.subcommand.replace('-', '_')))
+        func = getattr(am_client, '{0}'.format(
+            args.subcommand.replace('-', '_')))
+        if func:
+            res = func()
+            if type(res) is dict:
```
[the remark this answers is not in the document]
ross-spencer: :+1:

tests/test_amclient.py
```diff
@@ -24,6 +24,7 @@
 
 
 class TmpDir:
+
```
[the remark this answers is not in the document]
ross-spencer: This is showing up as a flake8 error locally.

ross-spencer:
Good stuff. I have committed an initial tranche of changes tonight. Will look at adding setuptools functionality in the morning. I think it sounds like a great change for folks who might want to work with this script.

Issue #51, opened 2018-03-07
Problem: amclient does not use log file location specified by `--log-file`


hakamine:
I tried to run the amclient script with the following bash script
```
# cat /etc/archivematica/automation-tools/amclient-close-completed-transfers.sh
#!/bin/bash
cd /usr/lib/archivematica/automation-tools/
/usr/share/python/automation-tools/bin/python -m transfers.amclient --log-file /var/log/archivematica/automation-tools/amclient.log  close-completed-transfers --am-user-name autobot <API_KEY>
```
(note that /usr/lib/archivematica/automation-tools is a symlink to /opt/archivematica/automation-tools)

However getting the following error:
```
# sudo -u archivematica /etc/archivematica/automation-tools/amclient-close-completed-transfers.sh
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/opt/archivematica/automation-tools/transfers/amclient.py", line 33, in <module>
    from transfers import utils
  File "transfers/utils.py", line 17, in <module>
    LOGGER = get_logger(defaults.AMCLIENT_LOG_FILE, defaults.DEFAULT_LOG_LEVEL)
  File "transfers/utils.py", line 13, in get_logger
    return loggingconfig.setup(log_level, log_file_name, "amclient")
  File "transfers/loggingconfig.py", line 54, in setup
    logging.config.dictConfig(CONFIG)
  File "/usr/lib/python2.7/logging/config.py", line 794, in dictConfig
    dictConfigClass(config).configure()
  File "/usr/lib/python2.7/logging/config.py", line 576, in configure
    '%r: %s' % (name, e))
ValueError: Unable to configure handler 'file': [Errno 13] Permission denied: '/opt/archivematica/automation-tools/transfers/amclient.log'
```
So it seems that it is still using the default log file location `(os.path.join(THIS_DIR, 'amclient.log'))`

hakamine:
BTW this problem also affects transfers.py, when transfers.py is run it tries to create a log file in path `(os.path.join(THIS_DIR, 'amclient.log'))`

ross-spencer:
@hakamine I have created a new issue that we can use to attach more substantial refactoring work to. 

[...]

PR #53, opened 2018-03-15
Add transfer_async

sevein:
`transfer_async.py` is an entry point that reuses what's in `transfer.py` but hijacks the only function that needed to be changed (`start_transfer`) so the script works against the new `/api/v2beta/package` API.

`transfer_async.py` does not run `pre-transfer` scripts because the transfers are started automatically so there is no room for that. In the future, `/api/v2beta/package` may learn a new `open` status that may allow us to solve this limitation.

This can be considered a walking skeleton. The plan is to eventually have it replace `transfer.py`. The advantage is that `/api/v2beta/package` does not block. This was a priority @ JiscRDSS because you can't have a connection open for hours like we do when we're copying big transfers - `/api/v2beta/package` solves that problem.

I've been testing in my Compose environment and the following command:

```
python -m transfers.transfer_async \
    --log-level DEBUG \
    --user test --api-key test --am-url http://127.0.0.1:62080 \
    --ss-user test --ss-api-key test --ss-url http://127.0.0.1:62081 \
    --transfer-source "b002cea4-82cf-4bc5-b4e5-fbc87443540d" \
    --transfer-path /home/test \
    --transfer-type standard
```

Issue: https://jiscdev.atlassian.net/browse/RDSSARK-480.
It depends on https://github.com/artefactual/archivematica/pull/936.

### DIFF
--- a/transfers/transfer_async.py
+++ b/transfers/transfer_async.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Automate Transfers.
+
+Helper script to automate running transfers through Archivematica.
+
+Similar to ``transfers.transfer`` but using the new `/api/v2beta` API when
+possible.
+"""
+
+from __future__ import print_function, unicode_literals
+
+import base64
+import os
+import sys
+
+import requests
+
+# Allow execution as an executable and the script to be run at package level
+# by ensuring that it can see itself.
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from transfers import transfer
+from transfers.loggingconfig import set_log_level
+from transfers.models import Unit
+from transfers.transferargs import get_parser
+from transfers.transfer import (
+    LOGGER, get_next_transfer, get_accession_id, main
+)
+from transfers.utils import fsencode
+
+
+class DashboardAPIError(Exception):
+    """Dashboard API error."""
+
+
+def _api_create_package(am_url, am_user, am_api_key,
+                        name, type, accession,
+                        ts_location_uuid, ts_path):
+    url = am_url + '/api/v2beta/package/'
+    headers = {
+        'Authorization': 'ApiKey {}:{}'.format(am_user, am_api_key)
+    }
+    data = {
+        'name': name,
+        'type': type,
+        'accession': accession,
+        'path': base64.b64encode(fsencode(ts_location_uuid) + b':' + ts_path),
+    }
+    LOGGER.debug('URL: %s; Headers: %s, Data: %s', url, headers, data)
+    response = requests.post(url, headers=headers, json=data)
+    response.raise_for_status()
+    LOGGER.debug('Response: %s', response)
+    resp_json = response.json()
+    error = resp_json.get('error')
+    if error:
+        raise DashboardAPIError(error)
+    return resp_json
+
+
+def _start_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, ts_path,
+                    depth, am_url, am_user, am_api_key, transfer_type,
+                    see_files, session, config_file):
+    """
+    Start a new transfer.
+
+    This is an early implementation and an alternative to
+    ``transfer.start_transfer`` that uses the new ``/api/v2beta/package`` API.
+    Because transfers start immediately we don't run the scripts inside the
+    pre-transfer directory like ``transfer.start_transfer`` does.
+
+    :param ss_url: URL of the Storage Sevice to query
+    :param ss_user: User on the Storage Service for authentication
+    :param ss_api_key: API key for user on the Storage Service for
+                       authentication
+    :param ts_location_uuid: UUID of the transfer source Location
+    :param ts_path: Relative path inside the Location to work with.
+    :param depth: Depth relative to ts_path to create a transfer from. Should
+                  be 1 or greater.
+    :param am_url: URL of Archivematica pipeline to start transfer on
+    :param am_user: User on Archivematica for authentication
+    :param am_api_key: API key for user on Archivematica for authentication
+    :param bool see_files: If true, start transfers from files as well as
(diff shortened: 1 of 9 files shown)


sevein:
@ross-spencer maybe you can take a quick look to this PR? It adds a new experimental `transfer_async.py` that we're going to be using @ JiscRDSS so it doesn't really change existing workflows. `transfer_async.py` may be evolving as we update the `/api/v2beta/package` API.

ross-spencer:
@sevein Acknowledging I still need to look at this! - One branch relevant to this is, https://github.com/artefactual/automation-tools/commit/76a8b61b5267eed234fd5ce72ff52f3dcf335232 I think we should get this reviewed and merged and then rebase your PR here to fit - what do you think? I'll check with @hakamine if he has any more info on issue #51 but I think this change is independent of that, whatever is causing #51 can be handled in another branch.

sevein:
> I think we should get this reviewed and merged and then rebase your PR here to fit - what do you think?

Sure. Ideally we wouldn't block paying work with minor stuff but I can understand the situation.

sevein:
Rebased.

transfers/transfer_async.py
```diff
@@ -0,0 +1,154 @@
...
+    }
+    LOGGER.debug('URL: %s; Headers: %s, Data: %s', url, headers, data)
+    response = requests.post(url, headers=headers, json=data)
+    response.raise_for_status()
+    LOGGER.debug('Response: %s', response)
+    resp_json = response.json()
+    error = resp_json.get('error')
+    if error:
+        raise DashboardAPIError(error)
+    return resp_json
```
ross-spencer: I would consider adding functionality from here to the `utils.py` script where applicable. This is very rough code here as I work out how to replace the current call, but I've been finding setting the headers for authentication more reliable (I don't think auth in a JSON body is supported by all our endpoints?) https://github.com/artefactual/automation-tools/blob/dev/client-reingest-compressed/transfers/utils.py#L64

But, my original issue https://github.com/artefactual/automation-tools/issues/52 may also have handled what you're doing here, grabbing the response details and being able to access them possibly via member vars, and not just having to rely on logging inside `utils.py`.

transfers/utils.py
```diff
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
 
+"""Where you put stuff when you can't think of a good name for a module."""
```
ross-spencer: LOL!

transfers/transferargs.py
```diff
@@ -0,0 +1,73 @@
...
+
+def get_parser(doc):
+    # Variable for conformance to flake8 line lenght below.
+    rawformatter = argparse.RawDescriptionHelpFormatter
+
+    parser = argparse.ArgumentParser(description=doc,
+                                     formatter_class=rawformatter)
+    parser.add_argument('-u', '--user', metavar='USERNAME', required=True,
+                        help='Username of the Archivematica dashboard user '
+                              'to authenticate as.')
```
ross-spencer: Indentation (pylint)

transfers/transferargs.py
```diff
@@ -0,0 +1,73 @@
...
+                                            'user.')
+    parser.add_argument(
+        '-t', '--transfer-source', metavar='UUID', required=True,
+        help='Transfer Source Location UUID to fetch transfers from.')
+    parser.add_argument(
+        # default=b'' to convert to bytes from unicode str provided by
+        # command line.
+        '--transfer-path', metavar='PATH', help='Relative path within the '
+                                                'Transfer Source. Default: ""',
+                                                type=fsencode, default=b'')
```
ross-spencer: Indentation.

transfers/transfer_async.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def _start_transfer(ss_url, ss_user, ss_api_key, ts_location_uuid, ts_path,
+                    depth, am_url, am_user, am_api_key, transfer_type,
+                    see_files, session, config_file):
+    """
```
ross-spencer: One thing I forgot. I wonder if there is an opportunity with this PR to consider the use of am_ss_client object to hold some of these values. There are a lot parameters to pass around. It would be nice to set  them once and early. That might open up more options, for example, enabling that object to initialise itself, e.g. from `ENV` vars in the OS. But, for now, it might be nice to just have a single object for the `am_` `ss_` parts of each new function call. 

What do you think?

transfers/transfer_async.py
sevein: I agree. Right now this is just doing things in the same way as it's done in `transfers.transfer.start_transfer` but I think we can get to your idea soon.

transfers/transfer_async.py
```diff
@@ -0,0 +1,154 @@
...
+    except (requests.exceptions.HTTPError,
+            ValueError, DashboardAPIError) as err:
+        LOGGER.error('Unable to start transfer: %s', err)
+        new_transfer = Unit(
+            path=target, unit_type='transfer', status='FAILED', current=False)
+        session.add(new_transfer)
+        return None
+
+    # We don't run the scripts in the pre-transfer directory because the
+    # _api_create_package automatically starts the transfer.
```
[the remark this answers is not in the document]
sevein: Ok done!

transfers/transfer_async.py
```diff
@@ -0,0 +1,154 @@
...
+)
+from transfers.utils import fsencode
+
+
+class DashboardAPIError(Exception):
+    """Dashboard API error."""
+
+
+def _api_create_package(am_url, am_user, am_api_key,
+                        name, type, accession,
```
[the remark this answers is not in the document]
sevein: Thanks, good catch. I have to tweak my Sublime a little more.

transfers/transfer_async.py
ross-spencer: I see, we're just trying to get to the point that `transfers.py` is at first and build from there. Works for me.

PR #54, opened 2018-03-20
Allow a logger to be passed to utils as an arg

### LINKED ISSUE #51: Problem: amclient does not use log file location specified by `--log-file`

### LINKED ISSUE #52: Problem: Turn transfers/utils.py into a self-contained library/class
The fix for #51 will likely be a workaround to get it up and running in good time. Currently, `utils.py` is called by two separate module scripts but it contains its own logging. 

It seems more appropriate for the scripts calling `utils.py` to manage their own logging and be able to retrieve what they need to report on from member variables within utils.py by turning it into one or many classes, e.g.

Instead of `utils.py` logging, `LOGGER.debug(response['content-type']), it should store the response in a member var. On returning an exception `amclient.py` or `transfers.py` can opt to look into the response attributes on their own terms. 

To improve this component of the transfers module, we will need to:

- Remove logging calls
- Raise exceptions instead of returning error codes when something goes wrong
- Assign components of the request and response to member variables where required
- Decide on which getters/setters will be most appropriate

This work could probably do with some discussion before it is implemented however.

### LINKED ISSUE #55: Problem: python shebang in py modules that are not executable
We only need to have a shebang if the script is an executable / entry point.

699b018046afba3aaac48356adfb2e2a70b2a075 introduced shebags in modules where I believe it's not needed.

```
> transfers/amclientargs.py:1:#!/usr/bin/env python
> transfers/defaults.py:1:#!/usr/bin/env python
> transfers/errors.py:1:#!/usr/bin/env python
> transfers/loggingconfig.py:1:#!/usr/bin/env python
> transfers/utils.py:1:#!/usr/bin/env python
```

ross-spencer:
The refactor in 699b018 didn't fully take into account the how logging
might work across the module. Python provides module level thread safe
logging at the module level through the logging module with appropriate
configuration.

This fix provides that functionality and additionaly sets a delay write
flag to ensure that no logs are created when there is no logging information
output.

Fixes #51
Resolves #52

Shebangs have been removed from non-executable scripts to resolve #55

Minor pylint changes added. Importantly, the incorrect check against a
non-existent ERR_AMCLIENT_UNKNOWN has been replaced with ERR_CLIENT_UNKNOWN.

--

Thanks to @sevein for his guidance on how to use the logging module effectively.

### DIFF
--- a/transfers/amclient.py
+++ b/transfers/amclient.py
@@ -13,6 +13,7 @@ import binascii
 import base64
 from collections import defaultdict
 import json
+import logging
 import os
 import pprint
 import re
@@ -32,13 +33,7 @@ from transfers import amclientargs
 from transfers import errors
 from transfers import utils
 
-
-def get_logger(log_file_name, log_level):
-    return loggingconfig.setup(log_level, log_file_name, "amclient")
-
-
-# Default logging if no other logging is provided in the class.
-LOGGER = get_logger(defaults.AMCLIENT_LOG_FILE, defaults.DEFAULT_LOG_LEVEL)
+LOGGER = logging.getLogger('transfers')
 
 
 def b64decode_ts_location_browse(result):
@@ -91,7 +86,7 @@ def is_uuid(thing):
     return defaults.UUID_PATT.search(thing) is not None
 
 
-class AMClient:
+class AMClient(object):
 
     def __init__(self, **kwargs):
         """Construct an Archivematica client. Provide any of the following
@@ -130,12 +125,13 @@ class AMClient:
                 # Shortening variable for PEP8 conformance.
                 err_lookup = errors.error_lookup
                 if isinstance(res, int):
-                    self.stdout(err_lookup.get(res,
-                                err_lookup(errors.ERR_AMCLIENT_UNKNOWN)))
+                    self.stdout(err_lookup
+                                .get(res,
+                                     err_lookup(errors.ERR_CLIENT_UNKNOWN)))
                 else:
                     self.stdout(getattr(self, method)())
             except:
-                self.stdout(errors.error_lookup(errors.ERR_AMCLIENT_UNKNOWN))
+                self.stdout(errors.error_lookup(errors.ERR_CLIENT_UNKNOWN))
         else:
             raise AttributeError('AMClient has no method {0}'.format(name))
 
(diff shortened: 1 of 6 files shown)


sevein:
> @sevein I did some further reading and grokked the module level logging idea using logging.getLogger() but I couldn't make it work for the way we have configured this module. The result No handlers could be found for logger "root" I'm afraid I couldn't work it out, so pointers on that would be helpful for this PR.

`logging.getLogger()` returns the root logger of the hierarchy which you had not defined in your dict config (see [schema details](https://docs.python.org/2/library/logging.config.html#configuration-dictionary-schema)).

I was thinking on something simpler like this: https://github.com/artefactual/automation-tools/compare/dev/revert-logging-refactor?expand=1. It brings back the old pattern we used to follow before your refactor excepting that it takes advantage of your loggingconfig module so we have the setup function defined only once, but the function doesn't return the logger it just updates the global config in the standard library.

I'm hoping that my branch can fix #51 and #52?

ross-spencer:
@sevein I've made the changes based on your branch. 

@hakamine I am pretty sure this does resolve the issue you were seeing now, because of my previous refactor the logger was still trying to write an empty log file to the default location. It won't happen with this new configuration. It wouldn't have happened previously either i believe because each script had been responsible for its own logging config and wasn't using a shared set of functions.

sevein:
👏 👏 👏

PR #57, opened 2018-04-09
Adds AM Client Capability in support of Reingest

ross-spencer:
This PR adds capability to AM Client in support of the API functions
required for automated reingest.

* Adds new functions to AM Client, including reingest and transfer status.
* Provides unit testing for the additional function calls.
* Extends utils.py request mechanism to enable http authentication headers
  to be set.
* A formatting issue was discovered in models.py while testing which
  is now corrected.
* Warning for non-JSON responses are only output if there is a non-OK
  HTTP response where responses from the API are not always JSON.

Submitted as part of Redmine ticket 11686 for CCA in support of automated reingest, cc. @timothyryanwalsh.

### DIFF
--- a/tests/test_amclient.py
+++ b/tests/test_amclient.py
@@ -3,6 +3,7 @@
     $ python -m unittest tests.test_amclient
 
 """
+import collections
 import os
 import shutil
 import unittest
@@ -69,9 +70,9 @@ class TestAMClient(unittest.TestCase):
         """Test closing completed transfers when there are completed transfers
         to close.
         """
-        response = amclient.AMClient(
-            am_api_key=AM_API_KEY, am_user_name=AM_USER_NAME,
-            am_url=AM_URL).close_completed_transfers()
+        response = amclient.AMClient(am_api_key=AM_API_KEY,
+                                     am_user_name=AM_USER_NAME,
+                                     am_url=AM_URL).close_completed_transfers()
         close_succeeded = response['close_succeeded']
         completed_transfers = response['completed_transfers']
         assert close_succeeded == completed_transfers
@@ -406,6 +407,38 @@ class TestAMClient(unittest.TestCase):
         for item in close_succeeded:
             assert amclient.is_uuid(item)
 
+    @vcr.use_cassette('fixtures/vcr_cassettes/test_hide_units.yaml')
+    def test_hide_units(self):
+        """Test the hiding of a unit type (transfer or ingest) via the
+        Archivematica API.
+        """
+        Result = collections.namedtuple('Result',
+                                        'uuid unit_type expected data_type')
+        hide_tests = [Result(uuid="fdf1f7d4-7b0e-46d7-a1cc-e1851f8b92ed",
+                             unit_type='transfer',
+                             expected={'removed': True},
+                             data_type=dict),
+                      Result(uuid="777a9d9e-baad-f00d-8c7e-00b75773672d",
+                             unit_type='transfer',
+                             expected=errors.ERR_INVALID_RESPONSE,
+                             data_type=int),
+                      Result(uuid="b72afa68-9e82-410d-9235-02fa10512e14",
+                             unit_type='ingest',
+                             expected={'removed': True},
+                             data_type=dict),
+                      Result(uuid="777a9d9e-baad-f00d-8c7e-00b75773672d",
+                             unit_type='ingest',
+                             expected=errors.ERR_INVALID_RESPONSE,
+                             data_type=int),
+                      ]
+        for test in hide_tests:
+            response = amclient.AMClient(
+                am_api_key=AM_API_KEY,
+                am_user_name=AM_USER_NAME,
+                am_url=AM_URL).hide_unit(test.uuid, test.unit_type)
+            assert isinstance(response, test.data_type)
+            assert response == test.expected
+
     @vcr.use_cassette(
         'fixtures/vcr_cassettes/completed_ingests_no_ingests.yaml')
     def test_completed_ingests_no_ingests(self):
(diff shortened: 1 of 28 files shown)


ross-spencer:
@sevein @jraddaoui @jrwdunham I have added a second commit to include some additional functions which will help support re-ingest. I'll squash those following review. :+1:

transfers/utils.py
```diff
@@ -40,9 +51,14 @@ def _call_url_json(url, params, method='GET'):
...
             return response.json()
         except ValueError:  # JSON could not be decoded
-            LOGGER.warning('Could not parse JSON from response: %s',
-                           response.text)
-            return errors.ERR_PARSE_JSON
+            if not response.ok:
+                LOGGER.warning('Could not parse JSON from response: %s',
+                               response.text)
+                return errors.ERR_PARSE_JSON
+            pass
```
jrwdunham: We are already checking `if not response.ok` on line 46 so I don't think the block in 55-57 can ever be reached. 

Also, it is confusing (though not introduced by this PR) that we have nested `try/except` blocks. Can we separate out these different error conditions?

I think what you're trying to do here is allow for a non-JSON response. So maybe it would be better to have a `json` (or `assume_json`) kwarg in the function that defaults to `True` but when `False` returns `response.text`. How about something like the following?:

    def _call_url_json(url, params, method=METHOD_GET, headers=None,
                    assume_json=True):
        """Helper to GET a URL where the expected response is 200 with JSON.

        :param str url: URL to call
        :param dict params: Params to pass as HTTP query string or JSON body
        :param str method: HTTP method (e.g., 'GET')
        :param dict headers: HTTP headers
        :param bool assume_json: set to False if the response body should not be decoded as JSON
        :returns: Dict of the returned JSON or an integer error
                code to be looked up
        """
        method = method.upper()
        LOGGER.debug('URL: %s; params: %s; method: %s', url, params, method)
        try:
            if method == METHOD_GET:
                response = requests.request(method, url=url, params=params,
                                            headers=headers)
            else:
                response = requests.request(method, url=url, data=params,
                                            headers=headers)
                                            headers=headers)
            LOGGER.debug('Response: %s', response)
            LOGGER.debug('type(response.text): %s ', type(response.text))
            LOGGER.debug('Response content-type: %s',
                        response.headers['content-type'])
        except (urllib3.exceptions.NewConnectionError,
                requests.exceptions.ConnectionError) as err:
            LOGGER.error("Connection error %s", err)
            return errors.ERR_SERVER_CONN
        if not response.ok:
            LOGGER.warning('%s Request to %s returned %s %s', method, url,
                        response.status_code, response.reason)
            LOGGER.debug('Response: %s', response.text)
            return errors.ERR_INVALID_RESPONSE
        if assume_json:
            try:
                return response.json()
            except ValueError:  # JSON could not be decoded
                LOGGER.warning('Could not parse JSON from response: %s',
                            response.text)
                return errors.ERR_PARSE_JSON
        return response.text

tests/test_amclient.py
```diff
@@ -502,6 +502,35 @@ def test_get_transfer_status_invalid_uuid(self):
...
+        message = response['message']
+        message_type = response["type"]
+        assert message == ("Fetched status for 23129471-09e3-467e-"
+                           "88b6-eb4714afb5ac successfully.")
+        assert message_type == "SIP"
+
+    @vcr.use_cassette('fixtures/vcr_cassettes/ingest_status_invalid_uuid.yaml')
+    def test_get_ingest_status_invalid_uuid(self):
+        """Test the successful return of the status of an ingest for a
+        valid SIP UUID.
```
jrwdunham: I think this docstring is inaccurate. Aren't you testing for an invalid SIP UUID here?

transfers/amclient.py
```diff
@@ -336,6 +355,80 @@ def download_package(self, uuid):
...
+        returned which needs to be handled via error_lookup.
+        """
+        return utils._call_url_json(
+            '{0}/api/processing-configuration/{1}'
+            .format(self.am_url,
+                    processing_name),
+            params=None,
+            headers=self._am_auth_headers())
+
+    def approve_transfer(self, directory, transfer_type=TT_STANDARD):
```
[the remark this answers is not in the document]
ross-spencer: In the process of looking at this I did uncover a small bug in approve_transfer so I have completed a re-factor of that function which I'll submit as a separate commit in addressing the other CR points too. It can now use AMClient.

transfers/amclient.py
```diff
@@ -336,6 +355,80 @@ def download_package(self, uuid):
...
+        call.
+        """
+        url = '{0}/api/transfer/approve/'.format(self.am_url)
+        params = {"type": transfer_type, "directory": directory}
+        return utils._call_url_json(url,
+                                    headers=self._am_auth_headers(),
+                                    params=params,
+                                    method="POST")
+
+    def reingest_aip(self, pipeline_uuid, aip_uuid, reingest_type="full",
```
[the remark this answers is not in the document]
ross-spencer: Good knowledge! I hadn't seen this PR. The main function Holly added that isn't in AMClient yet is https://github.com/artefactual/automation-tools/blob/31a186887fc8886446604e7791710a2bbd21de7f/reingest/reingest.py#L140 `metadata()` which adds metadata to a SIP. Holly didn't write any tests for this, so I think for now we should close this PR and then consider how we might use it if we get a decent piece of work to develop and test against rather than trying to put it in now.

transfers/utils.py
```diff
@@ -26,7 +30,14 @@ def _call_url_json(url, params, method='GET'):
...
     try:
-        response = requests.request(method, url=url, params=params)
+        if headers is None:
+            response = requests.request(method, url=url, params=params)
+        else:
+            response = requests.request(method,
+                                        url=url,
+                                        data=params,
+                                        headers=headers)
+
```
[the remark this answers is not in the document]
ross-spencer: @jrwdunham I think you're spot on - that being said - making that change now means re-recording all the `vcrpy` tapes for `amclient.py` and 'transfer.py` I can log this as an issue to be fixed as I think preparing each of those combinations right now might take a good while.

transfers/utils.py
ross-spencer: Aside from the overhead of re-recording our `vcrpy` fixtures noted above, with slight modification I've added it into `utils.py` thanks for taking the time to write this version.

transfers/amclient.py
```diff
@@ -336,6 +355,80 @@ def download_package(self, uuid):
...
+        """
+        params = {'pipeline': pipeline_uuid,
+                  'reingest_type': reingest_type,
+                  'processing_config': processing_config}
+        url = "{0}/api/v2/file/{1}/reingest/".format(self.ss_url, aip_uuid)
+        return utils._call_url_json(url,
+                                    headers=self._ss_auth_headers(),
+                                    params=json.dumps(params),
+                                    method=utils.METHOD_POST)
+
```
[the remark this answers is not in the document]
ross-spencer: Good spot. I think we've got six new methods across these commits, I'll be sure to add them before re-review. :+1:

ross-spencer:
@jrwdunham this is ready for further review I believe. Following our past reviews I've tried to separate a number of commits logically enough for you to see the differences - please let me know if this works okay, or if there is still some improvement i can make in this area. Hope this work is a bit closer to the mark. Thanks again for your help. My work on `amclient.py` has been useful to learn from.

transfers/utils.py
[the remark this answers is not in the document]
sevein: Why would you need to re-record the tapes if passing `None` has the same effect?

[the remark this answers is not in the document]
sevein: I'm not sure if I understand the logic. We'd be passing data in the request body only if headers are passed?

transfers/errors.py
```diff
@@ -3,13 +3,15 @@
...
+ERR_INVALID_URL = 4
 ERR_CLIENT_UNKNOWN = -1
 
 error_codes = {
     ERR_INVALID_RESPONSE: "Invalid response from server, check amclient log",
     ERR_PARSE_JSON: "Could not parse JSON resposne, check amclient log",
     ERR_SERVER_CONN: "Error connecting to the server, check amclient log",
-    ERR_CLIENT_UNKNOWN: "Unknown return from amclient, check logs."}
+    ERR_CLIENT_UNKNOWN: "Unknown return from amclient, check logs",
+    ERR_INVALID_URL: "Invalid URL passed to AM Client"}
```
sevein: You may want to move the closing bracket to its own line and include the comma in the last item. It's a convention in Python that reduces risk and the size of diffs when the structure is edited.

sevein:
```python
error_codes = {
    ERR_INVALID_RESPONSE: "Invalid response from server, check amclient log",
    ERR_PARSE_JSON: "Could not parse JSON resposne, check amclient log",
    ERR_SERVER_CONN: "Error connecting to the server, check amclient log",
    ERR_CLIENT_UNKNOWN: "Unknown return from amclient, check logs",
    ERR_INVALID_URL: "Invalid URL passed to AM Client",
}
```

tests/test_amclient.py
```diff
@@ -436,6 +436,243 @@ def test_close_completed_ingests_no_ingests(self):
...
+        'fixtures/vcr_cassettes/transfer_status_invalid_uuid.yaml')
+    def test_get_transfer_status_invalid_uuid(self):
+        """Test the successful return of the status for a non-existant
+        transfer in Archivematica.
+        """
+        response = amclient.AMClient(
+            am_api_key=AM_API_KEY,
+            am_user_name=AM_USER_NAME,
+            am_url=AM_URL,
+            transfer_uuid="7bffc8f7-baad-f00d"
```
sevein: @jrwdunham this pattern where the client holds random attributes isn't very adequate for multi-threaded uses right?

transfers/utils.py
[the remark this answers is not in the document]
ross-spencer: @sevein just a mistake on my part. Additional unit tests added for `hide_unit` to help capture this again in future, and an analysis here: https://github.com/artefactual/automation-tools/issues/62 the comment primarily related to the proposed refactor below, but it was spot on, I was not. Solved now though.

transfers/errors.py
ross-spencer: :+1:

ross-spencer:
@sevein @jrwdunham hopefully the submitted commits pull this up to scratch now. I've tried to delineate commits as usefully as possible.

tests/test_amclient.py
```diff
@@ -436,6 +469,243 @@ def test_close_completed_ingests_no_ingests(self):
...
+        response = amclient.AMClient(ss_api_key=SS_API_KEY,
+                                     ss_user_name=SS_USER_NAME,
+                                     ss_url=SS_URL).get_pipelines()
+
+        objects = response['objects']
+        pipelines = objects[0]['uuid']
+        resource_uri = objects[0]['resource_uri']
+        assert amclient.is_uuid(pipelines)
+        assert resource_uri == ('/api/v2/pipeline/f914af05-c7d2'
+                                '-4611-b2eb-61cd3426d9d2/')
```
sevein: Why are you using parenthesis here?

tests/test_amclient.py
```diff
@@ -69,9 +70,9 @@ def test_close_completed_transfers_transfers(self):
         """Test closing completed transfers when there are completed transfers
         to close.
         """
-        response = amclient.AMClient(
-            am_api_key=AM_API_KEY, am_user_name=AM_USER_NAME,
-            am_url=AM_URL).close_completed_transfers()
+        response = amclient.AMClient(am_api_key=AM_API_KEY,
+                                     am_user_name=AM_USER_NAME,
+                                     am_url=AM_URL).close_completed_transfers()
```
sevein: I have one more suggestion inspired by Django's subclass of `TestCase`. Should we have a `tearUp` method that initializes the client in `self.client`? It would remove a bunch of duplicated code.

    resp = self.client.close_completed_transfers)

Or in cases where you have to change some of the properties...

    self.client.transfer_uuid = "63fcc1b0-f83d-47e6-ac9d-a8f8d1fc2ab9"
    resp = self.client.get_transfer_status()

tests/test_amclient.py
ross-spencer: @sevein I will remove in favour of backslash syntax. Parenthesis just looked better.

tests/test_amclient.py
ross-spencer: @sevein I like this as a future enhancement but would be more comfortable making a new issue of this so that I can merge this work for the client waiting on it and to consider the possible solutions. Does this sound right? If `self.client.transfer_uuid` isn't set in a tear-up method when the class is initialised then `setattr(object, name, value)` needs to be called to set one after the fact - we can't just access the members if they don't exist?

ross-spencer:
@sevein I have left a comment about the potential `tearup` method. Otherwise addressed that final syntax issue.

@jrwdunham @sevein it seems I just need explicit approval to squash and merge now. Let me know though if there are any blockers.

tests/test_amclient.py
```diff
@@ -482,8 +482,8 @@ def test_get_pipelines(self):
         pipelines = objects[0]['uuid']
         resource_uri = objects[0]['resource_uri']
         assert amclient.is_uuid(pipelines)
-        assert resource_uri == ('/api/v2/pipeline/f914af05-c7d2'
-                                '-4611-b2eb-61cd3426d9d2/')
+        assert resource_uri == '/api/v2/pipeline/f914af05-c7d2' \
+                               '-4611-b2eb-61cd3426d9d2/'
```
sevein: Oh I had never realized that you need parenthesis to do that.
I'm used to see this working:

```python
print("foo"
      "bar")
```

You can do as you prefer, this is new to me.

ross-spencer:
@jrwdunham I have squashed my commits and rebased this against `master` now. From a process perspective I just want to make sure you're happy for me to merge into the master branch at this point?

jrwdunham:
@ross-spencer I'm just having another look. Sorry I've taken so long. I'll have a response by end of day today.

transfers/amclientargs.py
```diff
@@ -159,6 +205,48 @@
...
+    ),
+    SubCommand(
+        name='reingest-aip',
+        help='Initiate the reingest of an AIP from the storage service with a given UUID.',
+        args=(PIPELINE_UUID, AIP_UUID, SS_API_KEY),
+        opts=(REINGEST_TYPE, PROCESSING_CONFIG, SS_USER_NAME, SS_URL, OUTPUT_MODE)
+    ),
+    SubCommand(
+        name='get-package-details',
+        help='Retrieve details about a pakcage in the storage service with a given UUID.',
```
jrwdunham: Typo: "pakcage".

PR #44, opened 2018-01-19
Add option to delete source files on complete

timothyryanwalsh:
Hi, Artefactual! This is a small suggested addition to transfer.py that might help reduce some of the manual labor still necessary with an automation-tools pipeline - namely, deleting transfer source files after transfers have completed.

Issue #75, opened 2018-06-03
Problem: Add delete on completion pull-request


ross-spencer:
We have received a PR: https://github.com/artefactual/automation-tools/pull/44 to add deletion on completion to Automation tools and have decided it should be merged into master following any additional code-review or enhancements.

ross-spencer:
Closed via https://github.com/artefactual/automation-tools/pull/79.

PR #79, opened 2018-06-04
Add delete-on-complete option to automation tools master branch

ross-spencer:
From PR #44, we have brought that PR into the Artefactual hosted repository, squashed the previous commits, and made some final modifications via commit 442c49f. 

@timothyryanwalsh are you happy with the final changes here? 442c49f 

* Added a note to the deletion log that a failure  might be because of the script being run remotely
* Corrected an argument being passed to `AMClient()`
* Added some minor string formatting changes to the `get_status()` function while there was an opportunity to look at the code
* Made the `--delete` flag more verbose, now `--delete-on-complete`
* Used `.get()` for dictionary look-ups to avoid potential `KeyError`

Let me know! 

Once squished again we'll close PR #44 and #75 via here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The `transfers.py` script can be modified to adjust how automated transfers work
 * `--transfer-type TYPE`: Type of transfer to start. One of: 'standard' (default), 'unzipped bag', 'zipped bag', 'dspace'.
 * `--files`: If set, start transfers from files as well as folders.
 * `--hide`: If set, hides the Transfer and SIP once completed.
+* `--delete`: If set, delete transfer source files from watched directory once completed.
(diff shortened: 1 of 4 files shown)


timothyryanwalsh:
LGTM! Thanks, @ross-spencer!

ross-spencer:
Thanks for the contribution @timothyryanwalsh. That's all merged. I hope it is still useful somewhere down the road for you! 

FWIW, your use of `['key']` syntax for `dict` access was as you saw, entirely in-keeping with the rest of the script. Moving them all to `.get()` is something we'll have to do a bit over time or next big refactor but this was a nice opportunity to fix one function up a bit so thanks for that opportunity as well.

PR #88, opened 2018-10-26
Update the documentation around reingest.py

ross-spencer:
There isn't a lot of information to describe what reingest.py is
expected to do, and how it should be monitored. We provide this
information here.

For reviewers, just one file has changed `README.md`. 

You can use the diff output if you like, but the rendered text can be viewed in the branch here: https://github.com/artefactual/automation-tools/tree/dev/issue-11686-document-reingest#reingest

Connected to archivematica/issues#300

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,24 +552,167 @@ times, e.g. `-qq`
 Reingest
 --------
 
-The *transfers/reingest.py* script is a module and CLI that provides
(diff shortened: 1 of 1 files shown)


ablwr:
I'd like this to be added in! Has anything changed that would warrant consideration before review?

ross-spencer:
Nope! @ablwr it would be one less thing on my list if I could finally make appropriate CR changes and merge.

PR #96, opened 2019-02-01
Correct the behavior of Invalid transfer path

luisilara:
When we use the parameter --files with the transfers.py script  os.path.isdir returns  a false statement

I repaced it with os.path.exists and will also return True if there's a regular file with that name.

Connected to archivematica/issues#461


ross-spencer:
Hi @luisilara just a reminder: don't forget to merge!! Nice work on your first AM PR! 

:tada: :tada: :tada:

PR #98, opened 2019-03-20
[WIP] Add script for validating Avalon Media System manifest CSV

ablwr:
I've been working with the folks over at [Avalon Media System](www.avalonmediasystem.org) to develop an integrated, mostly-automated workflow for using Archivematica and Avalon together. This script is meant to review and validate their specific flavor of a "manifest file", required by Avalon and containing descriptive and hierarchical metadata about the files. (I deduced these rules from conversation and this page in their documentation: https://wiki.dlib.indiana.edu/display/VarVideo/Batch+Ingest+Package+Format).

This is a [WIP] because I'm not sure at this stage if this script should live here in automation-tools, or be part of a larger initiative akin to the request made in this issue: https://github.com/archivematica/Issues/issues/563 by @goetzk and with some discussion and ideas contributed from @helrond and @ross-spencer.

This could be a supported use case for the initial development of an API endpoint, something in which could initially work with this Avalon CSV validation case, but could grow to support other forms, as suggested in the above Issue's thread. 

Of course, the location and the general structure of the script would have to change.

### DIFF
--- a/transfers/pre-transfer/validate_avalon_manifest.py
+++ b/transfers/pre-transfer/validate_avalon_manifest.py
@@ -1,0 +1,251 @@
+#!/usr/bin/env python3
+"""
+Parse Avalon Media System manifest file and validate against
+rules prior to ingest into Archivematica.
+"""
+
+import argparse
+import collections
+import csv
+import os
+import sys
+
+
+class ManifestValidator(object):
+    def __init__(self, source_folder):
+        self.source_folder = source_folder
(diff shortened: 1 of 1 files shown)


ablwr:
Also, my local validator doesn't seem to agree with the validator here, so I'll have to work on that too.

ablwr:
This has been moved and will be resolved in https://github.com/artefactual/archivematica/pull/1400.

PR #99, opened 2019-03-22
Remove local amclient and let automation-tools use PyPi import

ross-spencer:
This work removes the local copy of amclient from the automation-tools script and drops-in the recently created PyPi version. It also makes naming conventions for vcrpy cassettes consistent so that the modules that they are recorded for are clear to the reader. 

Connected to archivematica/issues#544
Connected to archivematica/issues#601


transfers/transfer.py
```diff
@@ -27,8 +27,8 @@
 # by ensuring that it can see itself.
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
+from amclient import AMClient
```
sevein: I think this import should now be moved to the third-party import group.

```python
from amclient import AMClient
import requests
from six.moves import configparser
from sqlalchemy.orm.exc import NoResultFound
```

tests/test_aips_models.py
```diff
@@ -3,7 +3,7 @@
 import unittest
 
 from aips import models
-from tests.test_amclient import TmpDir
+from tests.tests_helpers import TmpDir
```
sevein: Maybe `tests.helpers` is a better name? Or `tests.tmpdir`?

It reminds me that at some point we should just move over the [`tmp_path`](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture) pytest fixture.

ross-spencer:
Thanks for the quick CR @sevein :+1: Moving onto Black changes next!

PR #100, opened 2019-03-24
Use black formatting

ross-spencer:
Connected to archivematica/issues#393

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,12 +32,6 @@ matrix:
     - python: "3.6"
       env: TOXENV=py36
 
-    - python: "2.7"
-      env: TOXENV=py27-flake8
-
-    - python: "3.6"
-      env: TOXENV=py36-flake8
-
     # Linters
 
     - python: "3.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 ---
 
 sudo: false
+dist: xenial
 
 language: python
 cache: pip
+
+branches:
+  only:
+    - master
 
 addons:
   apt:
@@ -32,3 +37,8 @@ matrix:
 
     - python: "3.6"
       env: TOXENV=py36-flake8
+
+    # Linters
+
+    - python: "3.7"
+      env: TOXENV=linting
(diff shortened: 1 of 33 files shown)


.travis.yml
```diff
@@ -32,3 +37,8 @@ matrix:
 
     - python: "3.6"
       env: TOXENV=py36-flake8
```
sevein: py36-flake8 and py27-flake8 can be removed now, i.e.:

```yml
    - python: "2.7"
      env: TOXENV=py27-flake8

    - python: "3.6"
      env: TOXENV=py36-flake8
```

tox.ini
```diff
@@ -17,7 +17,12 @@ commands = flake8 .
 
 [flake8]
 exclude = .git, .tox, __pycache__, old, build, dist
+application-import-names = flake8
```
sevein: `[testenv:py27-flake8]` and `[testenv:py36-flake8]` can be removed now.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = {py27,py36}{,-flake8}
+envlist = {py27,py36}{,-flake8}, linting
```
sevein:
```suggestion
envlist = py27, py36, linting
```

.travis.yml
ross-spencer: Hi @sevein did you notice when you run tox with that newer setting then Flake8 doesn't run locally? That's often beneficial to the developer before pushing. I'm not sure though, maybe it's doing something different, or my configuration needs changing?

tox.ini
ross-spencer: Hi @sevein did you notice when you run tox with this newer setting then Flake8 doesn't run locally? That's often beneficial to the developer before pushing. I'm not sure though, maybe it's doing something different, or my configuration needs changing?

sevein: I think that your `tox.ini` is missing the `linting` definition:

```ini
[testenv:linting]
basepython = python3
deps = pre-commit
commands = pre-commit run --all-files --show-diff-on-failure
```

Then:

```
$ tox -e linting
black....................................................................Passed
flake8...................................................................Failed
hookid: flake8
```

ross-spencer: :+1: that looks better now.

.travis.yml
sevein: This needs to go!

PR #104, opened 2019-04-02
Add installation commands/instructions for Docker

### LINKED ISSUE #87: Problem: Consider making the installation instructions a little more clear
While installing the automation tools, a person not as familiar with python, or does not know what virtualenv is may get hung up on step two of the installation instructions: `Create virtualenv /usr/share/python/automation-tools and pip install requirements there`

ablwr:
Somewhat resolve #87.

This may just be for all environments, but since I am on Docker, I wrote them for Docker and haven't checked into a current "standard" environment to give it a try yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Installing automation-tools in Docker
 Run each line below to install automation-tools in a Docker environment:
 
 ```bash
-sudo docker exec -it compose_archivematica-dashboard_1 /bin/bash 
+sudo docker exec -it compose_archivematica-dashboard_1 /bin/bash
 apt-get update
 mkdir /usr/lib/archivematica
 mkdir /usr/lib/archivematica/automation-tools
--- a/README.md
+++ b/README.md
@@ -47,45 +47,51 @@ Requirements
 Installation
 ------------
 
-* Checkout or link the code in this repo to
-`/usr/lib/archivematica/automation-tools`
-* Create virtualenv `/usr/share/python/automation-tools` and pip install
-requirements there
-* Create directories `/var/log/archivematica/automation-tools` and
-`/var/archivematica/automation-tools` owned by user `archivematica`, for
-log/database/PID files.
-* Create directory `/etc/archivematica/automation-tools` and add configuration
-files there. Files in the `etc/` directory of this repository can be used as an
-example (also see below for more about configuration)
+Follow each of the steps below to install automation-tools:
 
-Installing automation-tools in Docker
--------------------------------------
-
-Run each line below to install automation-tools in a Docker environment:
-
+Make the following directories
 ```bash
-sudo docker exec -it compose_archivematica-dashboard_1 /bin/bash
-apt-get update
-mkdir /usr/lib/archivematica
-mkdir /usr/lib/archivematica/automation-tools
-cd /usr/lib/archivematica/automation-tools/
-git clone https://github.com/artefactual/automation-tools.git .
+mkdir -p /usr/lib/archivematica/automation-tools
 mkdir /usr/share/python/automation-tools
-cd /usr/share/python/automation-tools
-virtualenv .env
-source .env/bin/activate
-pip install -r /usr/lib/archivematica/automation-tools/requirements.txt
-mkdir /var/log/archivematica
-mkdir /var/log/archivematica/automation-tools
+mkdir -p /var/log/archivematica/automation-tools
 mkdir /var/archivematica/automation-tools
-chown archivematica:archivematica /var/log/archivematica/automation-tools
-chown archivematica:archivematica /var/archivematica/automation-tools
 mkdir /etc/archivematica/automation-tools
-apt-get install p7zip-full
-apt-get install sudo
 ```
 
-You may also want/need to `apt-get install vim`.
+Change ownership of the directories to be owned by archivematica (group archivematica)
+
+```bash
+chown archivematica:archivematica /var/log/archivematica/automation-tools
+chown archivematica:archivematica /var/archivematica/automation-tools
+```
+
+Clone this automation-tools repository into your `usr/lib`
+
+```bash
+cd /usr/lib/archivematica/automation-tools/
+git clone https://github.com/artefactual/automation-tools.git .
+```
+
+Set up Python's virtual environment
+
+```bash
+cd /usr/share/python/automation-tools
+virtualenv venv
+source venv/bin/activate
+pip install -r /usr/lib/archivematica/automation-tools/requirements.txt
+```
+
+(Optional) Update and retrieve required packages
+
+```bash
+apt-get update
+apt-get install p7zip-full
+```
+
+You may also want/need to `apt-get install sudo && apt-get install vim`.
+
+To use automation-tools, `cp` the scripts you want to use into your
+`/etc/archivematica/automation-tools` directory.
 
 Automated transfers
 -------------------
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-[![Travis CI](https://travis-ci.org/artefactual/automation-tools.svg?branch=master)](https://travis-ci.org/artefactual/automation-tools)
+[![Travis
+CI](https://travis-ci.org/artefactual/automation-tools.svg?branch=master)](https://travis-ci.org/artefactual/automation-tools)
 
 Automation Tools
 ================
@@ -24,7 +25,8 @@ automate the processing of transfers in an Archivematica pipeline.
     - [pre-transfer hooks](#pre-transfer-hooks)
     - [user-input](#user-input)
   - [Logs](#logs)
-  - [Multiple automated transfer instances](#multiple-automated-transfer-instances)
+  - [Multiple automated transfer
+    instances](#multiple-automated-transfer-instances)
   - [`transfer_async.py`](#transfer_asyncpy)
   - [Tips for ingesting DSpace exports](#tips-for-ingesting-dspace-exports)
 - [DIP creation](#dip-creation)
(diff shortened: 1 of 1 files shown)


ablwr:
Actually, the above does as described, but doesn't actually work.

I don't know the purpose of creating  /usr/share/python/automation-tools and setting up virtualenv inside of it. I can only get atools running if I set up my virtualenv inside of /user/lib/archivematica/automation-tools. This is how I got it running previously, so I tried to do it "the right way" according to the docs, but no luck.

Am I misunderstanding how this is intended to be used?

README.md
```diff
@@ -58,6 +58,32 @@ log/database/PID files.
 files there. Files in the `etc/` directory of this repository can be used as an
 example (also see below for more about configuration)
 
+Installing automation-tools in Docker
```
ross-spencer:
```suggestion
Running automation-tools in Docker Compose
```

ross-spencer: I am torn here, could be a discussion in Slack with folks - the management of our environment is done via docker-compose and so it makes the commands intuitive in one sense: `docker-compose exec --user=root archivematica-dashboard bash` whereas, interacting with `docker` works more globally across an environment. If we cut down the steps below to docker-variant vs. generic installation steps above, perhaps notes can be made for both?

README.md
```diff
@@ -58,6 +58,32 @@ log/database/PID files.
...
+sudo docker exec -it compose_archivematica-dashboard_1 /bin/bash
+apt-get update
+mkdir /usr/lib/archivematica
+mkdir /usr/lib/archivematica/automation-tools
+cd /usr/lib/archivematica/automation-tools/
+git clone https://github.com/artefactual/automation-tools.git .
+mkdir /usr/share/python/automation-tools
+cd /usr/share/python/automation-tools
+virtualenv .env
+source .env/bin/activate
```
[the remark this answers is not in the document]
ablwr: Is this required? I use `.env` by default because a lot of other Python programmers use it, and the virtualenv docs use `venv`. Within Python it doesn't really matter what is used here, it's like personal preference / syntax choice. Does it matter within automation-tools for a reason I haven't deduced?

README.md
ablwr: I made this Docker-specific just because I don't really have time (none of us do) for sorting out the differences. But I can try, and if there are none, I think we should just have one generic installation guide. Most people know how to get into Docker without needing the command. I copy-pasted from my notes as I went along trying to install this on a fresh VM.

ablwr:
Thanks @ross-spencer for the review and the additional info.

I wasn't using automation-tools for kicking off a transfer, just for picking up the create_dip.py and using it (and another script) in conjunction with each other inside of the Archivematica environment. So I wonder if I need to add but cut out a lot of your instructions and instead keep it to the point of installation. 

So it seems... the thing that I was missing was copying things from the usr/lib subfolder to the /etc subfolder. I guess I don't know why it is made to be this way. Is it just because it is best practices to put cloned github repo libraries into usr/lib, and then move only what is needed over to /etc?

ablwr:
@ross-spencer Also I know you have a docs PR too! I will review that later this evening when I get a chance! It is long overdue.

ross-spencer:
@ablwr I think it's just a 'good-ish' practice. If you were to recommend directly working the files distributed in the GitHub repos then the docs would need to include more thorough information about how to then keep the Git repository up to date to take advantage of performance or reliability improvements. E.g. users would need info on copying and renaming the examples, or making use of rebase functions in Git.

This assertion 'most of us know' I feel is a difficult position when writing docs. Given the opportunity we should be as clear as possible for the lowest barrier of entry. So it would be good to keep that in mind as we flesh these out. I think the extra time wherever it needs to be put, will pay off - for those outside, internal or new to Artefactual too. 

I do have time to help co-author these if you need, CR them too of course. Just let me know.

README.md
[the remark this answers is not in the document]
ross-spencer: Not really, but it's probably a good idea  to isolate the automation tools' dependencies when using a container that's dedicated to something else. 

FWIW I don't tend to use `venvs` on my local machine so I get the personal preference thing.

ablwr:
@ross-spencer I added your great example and cleaned up a few things. Want to re-review? Or add anything else?

Also, my auto-linter went a litttttttle overboard... :grimacing:

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+```
+
+1. (Optional) Update and retrieve required packages
+
+```bash
+apt-get update
+apt-get install p7zip-full
+```
+
+You may also want/need to `apt-get install sudo && apt-get install vim`.
```
ross-spencer: Today I learned `sudo` itself can be installed like this!! 

Personally, I don't use `vim` but in general, I think here we need to expand on why a user might need to do this. `(For docker users)' might be one approach. 

We could also update the docker images as well. There are a few pieces missing from those components that might be useful to folk.

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+
+To use automation-tools, `cp` the scripts you want to use into your
+`/etc/archivematica/automation-tools` directory.
```
ross-spencer: Perhaps we can break up the text here with a further sub-heading:

`### Configuring the automation tools to run`

I'm not sure if that would be a H2 (##) or H3 (###) or what the resturcturedText version of that looks like. 

We can then rework this sentence to remove the abbreviated `cp` and replace it with copy, so that we follow the pattern you've established elsewhere to say: 'this is what you're going to do' followed by 'this is the command you're going to use to do that'. Which I feel is good practice to capture a broad reader-type. 

Perhaps line `100: Follow the steps below as an example for setting up a script.` can form part of an introductory paragraph to what the reader is doing next.

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+
+Follow the steps below as an example for setting up a script.
+
+1. Copy the configuration file `cp
```
ross-spencer: Missing two newlines.

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+   /usr/lib/archivematica/automation-tools/etc/transfers.conf
+   /etc/archivematica/automation-tools/`
+
+1. Copy the bash script
```
ross-spencer:
```suggestion
1. Copy the transfer bash script:
```

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+
+`cp /usr/lib/archivematica/automation-tools/etc/transfer-script.sh
+/etc/archivematica/automation-tools/`
+
+1. Create a transfer source for the automation-tools in the storage service
+
+You can use an existing transfer source, or set up a new one.
+
+1. Your default bash script should be configured along the lines of this:
```
ross-spencer:
```suggestion
1. Your default transfer bash script should be configured along the lines of this:
```

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+
+`transfer-script.sh`:
+
+```bash
+#!/bin/bash
+# docker transfer script example
```
ross-spencer:
```suggestion
```

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+ --ss-url http://archivematica-storage-service:8000 \
+ --user test \
+ --api-key test  \
+ --ss-user test \
+ --ss-api-key test \
+ --transfer-source <transfer-source-in-storage-service> \
+ --config-file transfers.conf
+```
+
+5. Run the shell script!
```
ross-spencer: We should be consistent with `bash` vs. `shell`. I prefer `shell` I was tempted to use that above in a suggestion, but one or the other is fine. Or just `Run the script` too.

README.md
```diff
@@ -47,30 +49,106 @@ Requirements
...
+
+`$ ./transfer-script.sh` and you should have success!
```
ross-spencer: I think we're missing a step {n} to `chmod +x` the script to add executable permissions before running.

ablwr:
The linter is still throwing warnings but it's for the parts of the README that I'm not touching, although it did change things that it was deeming as broken. But the linebreaks (numbering, headers) aren't considered breaking changes, just warnings.

This repo doesn't have a contributing guide, and README guidelines seem they'd be different from a documentation framework like Sphinx. The README is already inconsistent with applying headers as underline-dashes or using hash marks. So which is preferred?

README.md
ablwr: I will leave this out. The terminal will certainly let the user know if they don't have one of these things.

ablwr:
OK updated with changes!

I was using the 1., 1., 1. pattern because [markdownlint](https://github.com/DavidAnson/markdownlint/blob/v0.13.0/doc/Rules.md) was recommending it, but I reverted it back. Markdownlint thinks there are many errors but since we don't have a recommended style guide for README files, I am not going to make them.

ross-spencer:
Hi @ablwr my understanding is that the style should be consistent with [Archivematica Docs](https://github.com/artefactual/archivematica-docs/blob/12da09e3664025169450ac8b913639ff85ebba84/style_guide.md) but I agree, some clarification around markdown and README would be good. There are separate issues for a few repositories about adding these guides, it might be worth opening up a general issue for all repos to either point to the AM standard or have their own standard. 

Thanks for making the changes. I'll try and have look before COP today. :+1:

README.md
```diff
@@ -42,38 +44,120 @@ automate the processing of transfers in an Archivematica pipeline.
...
+ --api-key test  \
+ --ss-user test \
+ --ss-api-key test \
+ --transfer-source <transfer-source-in-storage-service> \
+ --config-file transfers.conf
+```
+
+5. Change permissions setting on the script:
+
+`chmod -x transfer-script.sh`
```
ross-spencer:
```suggestion
`chmod +x transfer-script.sh`
```

ablwr:
Whoops. Nice small catch!

PR #107, opened 2019-04-10
WIP Add create_avalon_dip script

ablwr:
This is connected to https://github.com/archivematica/Issues/issues/643. 

Explanation of what this does is in the issue above and also commented alongside.

For some explanation towards my choices for the purpose of code review:
I think if this was going into the main code base, I would have extended this out into smaller functions that group together. But for me, the ethos of this repository is for people to read and customize things as they see fit, and I think it's more straightforward to keep it in one lengthy function for the use-case of people wanting to understand how and what this script does but are maybe not proficient in Python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-[![Travis CI](https://travis-ci.org/artefactual/automation-tools.svg?branch=master)](https://travis-ci.org/artefactual/automation-tools)
+[![Travis
+CI](https://travis-ci.org/artefactual/automation-tools.svg?branch=master)](https://travis-ci.org/artefactual/automation-tools)
 
 Automation Tools
 ================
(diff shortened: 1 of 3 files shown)


ablwr:
@jraddaoui any chance you can counter-review this? It's only a slight modification of the create_dip.py script that you just updated (thank you).

ablwr:
Hey @jraddaoui -- that makes sense to me and was my preferred option, but when this script was written a month or two ago, it was unclear how much create_dip would be changing in light of the CCA work and within the context of other people using it that we don't know about (which speaks to some of my frustration about the automation-tools lack of semver, for example). I also initially planned to perform create_dip and then build a script that moves things around in the Avalon-requested structure, but it seemed like it was doing unnecessary work.

I'll ping you on Slack to talk about how this relates to two client projects, but thanks for reviewing and I appreciate all the comments! I will make the changes some sort of way when we come up with a solution.

ablwr:
I am going to close this PR based on the create_dip changes. I incorporated this script into the new create_dip and will open a subsequent PR.

PR #108, opened 2019-05-05
Create DIP: support PREMIS 2.2 and 3.0 namespaces

jraddaoui:
Upgrading `metsrw` was not required but it's the only script using it and it works in both versions.

https://github.com/archivematica/Issues/issues/653

### DIFF
--- a/aips/create_dip.py
+++ b/aips/create_dip.py
@@ -192,6 +192,8 @@ def create_dip(aip_dir, aip_uuid, output_dir):
     shutil.move(aip_mets_file, to_zip_mets_file)
 
     mets = metsrw.METSDocument.fromfile(to_zip_mets_file)
+    namespaces = metsrw.utils.NAMESPACES.copy()
+    premis_map = metsrw.plugins.premisrw.utils.PREMIS_VERSIONS_MAP
     fsentries = mets.all_files()
     for fsentry in fsentries:
         if fsentry.use != "original" or not fsentry.path or not fsentry.file_uuid:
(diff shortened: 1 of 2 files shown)


aips/create_dip.py
```diff
@@ -27,6 +27,11 @@
 THIS_DIR = os.path.abspath(os.path.dirname(__file__))
 LOGGER = logging.getLogger("create_dip")
 
+PREMIS_VERSIONS_NS = {
+    "2.2": "info:lc/xmlns/premis-v2",
+    "3.0": "http://www.loc.gov/premis/v3",
+}
```
sevein: You could use `metsrw.plugins.premisrw.utils.*` instead.

aips/create_dip.py
```diff
@@ -220,9 +226,15 @@ def create_dip(aip_dir, aip_uuid, output_dir):
...
-        )
+        # Update PREMIS namespace based on version attribute
+        premis_version = premis.get("version", "2.2")
+        try:
+            namespaces["premis"] = PREMIS_VERSIONS_NS[premis_version]
+        except KeyError:
+            LOGGER.warning(
+                "Could not update namespace for PREMIS version: %s" % premis_version
+            )
+        original_name = premis.findtext("premis:originalName", namespaces=namespaces)
```
sevein: You may want to use `premisrw` instead because it takes care of this for you. E.g.:

```python
>>> premis_objects = fsentry.get_premis_objects()
>>> premis_objects[0].object_characteristics_extension
```

jraddaoui: I tried that, but `original_name` is not represented in the PREMISObject schema. I was going to create an issue to add it in there, but we also need the last modified data from FITS in here.

aips/create_dip.py
jraddaoui: It was a little ugly to use `metsrw.plugins.premisrw.utils.`PREMIS_VERSIONS_MAP[premis_version]["namespaces"]["premis"]` to update the `metsrw.utils..NAMESPACES` copy, but I'll give it another look.

jraddaoui: Done, thanks!

ablwr:
yay!

PR #123, opened 2019-08-29
Added script to export DIP data into NetX.

mcantelon:
I'll add documentation and tests ASAP, but wanted to get this pull request under review as a client needs it at the start of next week.


dips/copy_to_netx.py
```diff
@@ -0,0 +1,298 @@
...
+        sip_uuid = uuid_from_dip_path(dip_path)
+
+        for object_file in os.listdir(objects_path):
+            # copy file to NetX directory
+            shutil.copyfile(
+                os.path.join(objects_path, object_file),
+                os.path.join(netx_objects_directory, object_file),
+            )
+
+            # write CSV row
```
jraddaoui: Minor detail, most of the comments in this file start with uppercase, except this three comments.

dips/copy_to_netx.py
```diff
@@ -0,0 +1,298 @@
...
+    netx_objects_directory,
+    object_id,
+    component_id,
+    delete_local_copy,
+):
+    # Copy DIP to the currently processing location path, do not use any of the
+    # existing watched directories as that may trigger other workflows.
+    at_dips_dir = os.path.join(
+        shared_directory, "watchedDirectories", "automationToolsCopyToNetX"
+    )
```
jraddaoui: Is there a need to use the shared directory? I'd try to avoid that if possible, maybe using a working or output directory param.

dips/copy_to_netx.py
```diff
@@ -0,0 +1,298 @@
...
+        component_id = parse_component_id_from_mets(
+            os.path.join(dip_path, mets_filename)
+        )
+
+    if object_id is None:
+        LOGGER.info("Parsing %s for object ID." % mets_filename)
+        object_id = parse_object_id_from_mets(os.path.join(dip_path, mets_filename))
+
+    # Make sure NetX directories exist
+    if not os.path.exists(netx_csv_directory):
```
jraddaoui: Should we do this check first in this function and avoid all the above code?

dips/copy_to_netx.py
```diff
@@ -0,0 +1,298 @@
...
+        LOGGER.error(
+            "The specified NetX CSV directory doesn't exist: %s" % netx_csv_directory
+        )
+        return 1
+
+    if not os.path.exists(netx_objects_directory):
```
jraddaoui: Same in here, I'd do this check first.

README.md
```diff
@@ -722,6 +722,53 @@ following subset of arguments:
 
 See [Getting API keys](#getting-api-keys)
 
+Exporting to NetX
+-----------------
+
+The *dips/copy_to_netx.py* script is a command line tool that takes a DIP
+directory in the `watchedDirectories/uploadDIP` subdirectory of the
```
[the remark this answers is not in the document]
mcantelon: Good point. I've corrected the language.

README.md
```diff
@@ -722,6 +722,53 @@ following subset of arguments:
...
+Archivematica shared directory, parses the contents, and passes the objects,
+and CSV data describing the objects, to NetX.
+
+The component ID recorded in the NetX CSV can be:
+
+1) Specified as a command line option (`--component-id`)
+2) Specified using the Dublin Core identifier
```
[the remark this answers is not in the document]
mcantelon: I've updated so this (and the other XML lookups) now use absolute paths.

mcantelon:
Thanks @jraddaoui ... I've updated the branch. Let me know if there are any issues. I'll rebase it to clean up the commit history before I merge (as I did some commits to fix formatting).

dips/copy_to_netx.py
```diff
@@ -0,0 +1,319 @@
...
+    }
+    """
+
+    identifier = mets_data.xpath(
+        "/mets:mets/mets:dmdSec/mets:mdWrap/mets:xmlData/dcterms:dublincore/dc:identifier",
+        namespaces=namespaces,
+    )
+
+    if len(identifier):
+        return identifier[0].text
```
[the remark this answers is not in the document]
mcantelon: Ah, didn't realize they all are at the same level in the METS data. I've added logic to check the structmap to find the right DC metadata within METS.

mcantelon:
Hi @jraddaoui ... I've updated the PR. Sorry about amending a commit last time... didn't realize I'd done that. Let me know if there are any issues... thanks!

README.md
```diff
@@ -723,7 +723,7 @@ following subset of arguments:
 
 See [Getting API keys](#getting-api-keys)
 
-DIP Export To NetX
+DIP export to NetX
```
jraddaoui: Thanks! If you could also change it in the TOC, that would be great.

dips/copy_to_netx.py
```diff
@@ -74,27 +81,38 @@ def parse_component_id_from_mets(mets_data, namespaces):
...
-    """
+    # First: attempt to find it as the SIP-level Dublin Core identifier
+    objects_div = mets_data.xpath(
+        "/mets:mets/mets:structMap/mets:div/mets:div[@LABEL='objects']",
+        namespaces=namespaces
+    )
+
+    if len(objects_div):
+        # Use DMD ID find correct Dublin Core identifier
+        dmd_id = objects_div[0].get('DMDID')
```
jraddaoui: The DMDID attribute can actually contain multiple DMD ids, it's not very intuitive but correct in terms of the standard. When multiple DMD sections are associated to the same element, you should try to find the latest one.

See how we fixed this in AtoM:

https://github.com/artefactual/atom/commit/4764110f9426d4fc288b2e64bd185a7fb3d0c053

And how it's being done in SCOPE (not really the best way as it queries for and orders all DMD sections and then filters, instead of only querying for the ones you want and sort):

https://github.com/CCA-Public/scope/blob/master/scope/parsemets.py#L301-L309

SCOPE makes sure that DC metadata exists in the DMD sections. This is to workaround an specific issue they had with missing metadata in the most recent section. I consider this a workaround because the updated metadata could actually be no metadata (intentionally removing the original metadata), but it may a good addition for this case too.

mcantelon:
Thanks @jraddaoui ! Would it be possible to leave the DMID to later in the week and merge what's here for now? Or does the DMID issue seem like it's likely to pop up in the short term?

dips/copy_to_netx.py
mcantelon: Was the CREATED attribute recently added in AM's METS files? I can't find it in the METS files on the AM testing instance I'm using.

jraddaoui: It's not added in the original DMD sections, just in the updated ones. You may need to trigger a partial re-ingest and change part of the Dublin Core metadata in the AM GUI to get those updated DMD sections.

mcantelon: I talked to Justin and he's okay with the script not having logic to get the component ID from Dublin Core so I've removed it to simplify things (they'll still be able to netx.json to specify the component ID, though).

mcantelon:
Hi @jraddaoui ... I've updated the PR to get rid of the tricky bit.

README.md
```diff
@@ -722,6 +723,54 @@ following subset of arguments:
...
+The *dips/copy_to_netx.py* script is a command line tool that takes a DIP
+directory, typically in the `watchedDirectories/uploadDIP` subdirectory of the
+Archivematica shared directory, and exports data to NetX. The script copies
+the objects in the DIP, and writes CSV data describing the objects, to
+directories that NetX imports from.
+
+The component ID recorded in the NetX CSV can be:
+
+1) Specified as a command line option (`--component-id`)
+2) Specified using the SIP-level Dublin Core identifier
```
jraddaoui: We should remove this now.

Issue #128, opened 2021-11-01
create_dip.py script changes file names from BagIt transfers


djjuhasz:
Running the create_dip.py on an AIP created from a transfer of a BagIt bag causes the initial three characters to be removed from all of the file names in the DIP objects directory.  The following warning is output when this occurs:

```
INFO 2021-08-03 15:36:40 Moving file: objects/box4b12.bin
WARNING 2021-08-03 15:36:40 premis:originalName not starting with %transferDirectory%objects/
Custom Fields
```

djjuhasz:
Fix merged: https://github.com/artefactual/automation-tools/commit/e5a9f8077dfa19c87897d2f41610636d4faaf9b8

PR #129, opened 2021-11-02
Fix premis:originalName parsing, refs #128

djjuhasz:
Use `os.path.basename()` to get file names from the
`<premis:originalName>` path.

The previously assumed path of `%transferDirectory%objects` is not valid
for BagIt transfers, which lead to truncating the first three characters
of the filename in the DIP objects zip file.

### DIFF
--- a/aips/create_dip.py
+++ b/aips/create_dip.py
@@ -240,13 +240,17 @@ def create_dip(aip_dir, aip_uuid, output_dir, mets_type, dip_type):
         premis = techmd.contents.document
         update_premis_ns(premis, namespaces, premis_map)
 
-        original_filename = get_original_filename(premis, namespaces)
-        if not original_filename:
+        original_name = get_premis_original_name(premis, namespaces)
+        if not original_name:
             LOGGER.warning("Could not get original file name from premis:originalName")
             continue
 
+        original_relpath = get_original_relpath(original_name)
+        if not original_relpath:
+            continue
+
         # Move original file with original file name and create parent folders
-        dip_file_path = os.path.join(to_zip_dir, original_filename)
+        dip_file_path = os.path.join(to_zip_dir, original_relpath)
         dip_dir_path = os.path.dirname(dip_file_path)
         if not os.path.exists(dip_dir_path):
             os.makedirs(dip_dir_path)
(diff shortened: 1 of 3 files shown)


tw4l:
@djjuhasz Based on the [paths expected in the tests](https://github.com/artefactual/automation-tools/blob/master/tests/test_create_dip.py#L93-L100), it looks like we don't always want just the basename for files in an AIP, but the full relative path from the root directory (which we're currently assuming is always `%transferDirectory%objects`, but may actually be `%transferDirectory%data` if I'm understanding correctly).

I'd look into either [`os.path.relpath`](https://docs.python.org/3/library/os.path.html#os.path.relpath) or check explicitly for the full strings we want to remove and if either is found, splice them from the beginning of the string.

djjuhasz:
@tw4l I've updated the code to return the relative paths from the root directory.

aips/create_dip.py
```diff
@@ -411,15 +415,25 @@ def update_premis_ns(premis, namespaces, premis_map):
...
         LOGGER.warning("premis:originalName could not be found")
-    string_start = "%transferDirectory%objects/"
-    if original_name[:27] != string_start:
-        LOGGER.warning("premis:originalName not starting with %s", string_start)
-    return original_name
+        return None
+
+    # Strip the path prefix from the file path, and return the remainder
+    path_prefixes = ["%transferDirectory%objects/", "%transferDirectory%data/"]
+    for pp in path_prefixes:
```
tw4l, line 428: Nit: We try to avoid single- (or double-) letter variable names. Could rename to `prefix`?

aips/create_dip.py
```diff
@@ -411,15 +415,25 @@ def update_premis_ns(premis, namespaces, premis_map):
...
+        if original_name[: len(pp)] == pp:
```
tw4l, line 429: Nit: this could be `if original_name.startswith(pp)`, which might be a bit more readable.

tw4l (approved):
Thanks @djjuhasz, this looks great! I've left a few in-line comments with nits that you can address if you'd like before merging. It would also be great to have a test case for the `"%transferDirectory%data/` path to codify this knowledge in the tests if you have the time.

aips/create_dip.py
djjuhasz: Haha, I thought I was being all Pythonic by using a two letter variable name - most of the examples in the docs use single letter vars for temporary values from an iterator.  I'll switch it up.

djjuhasz:
@tw4l third time is the charm? 😆

djjuhasz:
@tw4l nevermind, the unit tests failed in Python 2.7 😢

tw4l:
@djjuhasz It looks like unittest.assertLogs is new in Python 3.4, which would explain the failing test for 2.7: https://docs.python.org/3/library/unittest.html?highlight=unittest#unittest.TestCase.assertLogs. You could change the test case to verify that the function returns `None` for an invalid path?

djjuhasz:
@tw4l four to go? 😜

tw4l:
@djjuhasz LGTM!

No need to change this, but looking at the tests again it occurred to me that you might be interested in pytest's [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html) functionality.

It enables combining tests like these and testing different combinations of inputs and outputs, so that these three tests:

```python3
 def test_get_original_relpath_objects_dir(self):
    path = "%transferDirectory%objects/folder1/file5.txt"

    assert create_dip.get_original_relpath(path) == "folder1/file5.txt"

def test_get_original_relpath_data_dir(self):
    path = "%transferDirectory%data/folder1/file5.txt"

    assert create_dip.get_original_relpath(path) == "folder1/file5.txt"

def test_get_original_relpath_warn_invalid_prefix(self):
    path = "%transferDirectory%datas/folder1/file5.txt"

    assert create_dip.get_original_relpath(path) is None
```

can become, e.g.:

```python3
import pytest

...

@pytest.mark.parametrize(
    "path,expected_return",
    [
        # %transferDirectory%objects prefix
        (
            "%transferDirectory%objects/folder1/file5.txt",
            "folder1/file5.txt"
        ),
        # %transferDirectory%data prefix
        (
            "%transferDirectory%data/folder1/file5.txt",
            "folder1/file5.txt"
        ),
        # Invalid prefix
        (
            "%transferDirectory%datas/folder1/file5.txt"
            None
        ),
    ],
)
def test_get_original_relpath(path, expected_return):
    assert create_dip.get_original_relpath(path) == expected_return
```

sevein:
2ee73fa9cc43ab0b6d6395bc4aa8173215935c36 and e5a9f8077dfa19c87897d2f41610636d4faaf9b8

PR #130, opened 2022-06-06
Fix transfer_async.py and add ability to specify processing config by name

tw4l:
This pull request adds a `processingconfig` setting in the `transfers.conf` file that allows users to specify which processing configuration to use for transfers created with the `transfer_async.py` script.

`transfer_async.py` wasn't working following a refactor of `transfer.py`, so I've also fixed a few things (a character encoding issue, out-of-date function signature, and switched to using the existing helpers in the `models` module for database interactions) to enable this feature to work.

I also added an integration test for `transfer_async.py` to help prevent future breaking changes.

### DIFF
--- a/etc/transfers.conf
+++ b/etc/transfers.conf
@@ -1,8 +1,13 @@
 # automation-tools:transfers configuration file example
 # /etc/archivematica/automation-tools/transfers.conf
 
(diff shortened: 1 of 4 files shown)


djjuhasz (approved):
Looks good to me 👍

PR #133, opened 2023-02-03
Add required options to transfer script template.

mcantelon:
Added missing required command-line options (Archivematica and storage
service URLs) to transfer script tempate and added backslashes so
command-line options aren't treated as separate commands.

### DIFF
--- a/etc/transfer-script.sh
+++ b/etc/transfer-script.sh
@@ -2,10 +2,12 @@
 # transfer script example
 # /etc/archivematica/automation-tools/transfer-script.sh
 cd /usr/lib/archivematica/automation-tools/
-/usr/share/python/automation-tools/venv/bin/python -m transfers.transfer
-  --user <user> 
-  --api-key <apikey>  
-  --ss-user <user> 
-  --ss-api-key <apikey> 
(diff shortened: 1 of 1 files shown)


etc/transfer-script.sh
```diff
@@ -2,10 +2,12 @@
...
-  --ss-user <user> 
-  --ss-api-key <apikey> 
-  --transfer-source <transfer_source_uuid> 
+/usr/share/python/automation-tools/venv/bin/python -m transfers.transfer \
+  --am-url <archivematica_url> \
+  --ss-url <storage_service_url> \
+  --user <user> \
+  --api-key <apikey> \ 
+  --ss-user <user> \
+  --ss-api-key <apikey> \
```
replaceafill, line 11: Would it make sense to prefix these example placeholders such as `<archivematica_user>`, `<archivematica_apikey>`, `<storage_service_user>` and `<storage_service_apikey>` like the `url` ones?

replaceafill (approved):
Thanks @mcantelon!

etc/transfer-script.sh
mcantelon: Good call! I'll do that.

PR #144, opened 2023-12-11
Dev/contributing

eviau-artefactual:
We've noticed that there was no contributing.md file in this repo.

Here is a suggestion that refers to the main AM guidelines.

What do you think?

If this text makes sense to you, I will use it to make other PRs to AM related projects.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,6 @@
 
 Thank you for your interest in making a contribution to automation-tools.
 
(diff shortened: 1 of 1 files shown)


eviau-artefactual:
About linting: if my understanding is correct, the tool used doesn't do the linting for you... is that correct ?

replaceafill (approved):

replaceafill:
> About linting: if my understanding is correct, the tool used doesn't do the linting for you... is that correct ?

That's correct. We haven't set up the `fix` option in `markdownlint-cli` by default. I'd recommend you to set up `pre-commit` (`pre-commit install`) in our repos so it helps you catching linting problems before committing.

